PR #3, opened 2026-01-15
add session track brief

evalstate:
Add the Session Track briefing document to the repository.

## Motivation and Context

To generate a PR to collect first round of reviews and comments.

## Types of changes

- [ ] New feature (non-breaking change which adds functionality)
- [X] Documentation update

## Checklist

- [X] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [X] My code follows the repository's style guidelines
- [X] New and existing tests pass locally
- [X] I have added appropriate error handling
- [X] I have added or updated documentation as needed


proposals/session-track.md
```diff
@@ -0,0 +1,19 @@
+**Sessions Track: Brief**
+
+**Goal** To decouple session state from the underlying transport connection, allowing for robust, long-lived application state that survives connection drops and scales horizontally across stateless servers.
+
+**Core Principle: Adapted Cookie Semantics**
+
+* **RFC 6265 as a Foundation:** We will use **RFC 6265 (HTTP Cookies)** as a functional starting point to determine necessary adaptations for MCP.  
+* **Payload Integration:** Session data will primarily be included as a dedicated field within the JSON-RPC payload. We will look to **SEP-1655** and **SEP-1685** for inspiration on structuring this metadata.
```
LucaButBoring, line 8: The payload integration is how we'll support this on stdio, right? I believe that was the plan, but wanted to verify that, as most of the discussion is (rightly) focused on HTTP servers.

proposals/session-track.md
```diff
@@ -0,0 +1,19 @@
...
+
+**Key Requirements**
+
+* **Routing Compatibility & Size Constraints:** To allow load balancers to route requests based on session affinity, we may need to copy the session identifier/token into an HTTP header. This introduces strict **size constraints**; the session token must remain small enough to fit within standard header limits (typically \~4KB), prohibiting large state blobs from being stored directly in the token.
```
LucaButBoring, line 12: Should we impose a size limit here? This is dependent on the network infra and is typically configurable - it [looks like](https://stackoverflow.com/a/8623061) only Nginx has a 4KB limit by default. It would likely be inadvisable to have even a 4KB header, but I believe that server operators will generally be able to determine what reasonable limits look like for their particular setups.

proposals/session-track.md
```diff
@@ -0,0 +1,19 @@
...
+* **Security & Integrity:** Servers utilizing cookies must encrypt or sign the data to prevent client-side tampering. We will evaluate including a standard mechanism for signing/encryption directly in the spec.  
+* **Expert Review:** We will consult security experts to review the design for known vulnerabilities common in standard cookie implementations.
+
+**Key Design Decision: Lifecycle Management**
+
+* **Implicit vs. Explicit:** We must decide between allowing lazy initialization (Implicit) versus requiring a formal setup handshake (Explicit).  
+  * *Consideration:* The need to **copy or fork** session data (e.g., branching an agent's state) strongly suggests preferring an **Explicit** mechanism (e.g., `session/create`, `session/fork`) to ensure these operations are predictable and race-free.
```
LucaButBoring, line 21: I don't believe forking sessions needs its own explicit RPC method, but that depends on our contract for cookie formats. I would expect that if I (as a client) save the current cookie at each point in the conversation, I can simply roll back to an old conversation state and do something different, and if that implies a different change to the session state, I'll just wind up receiving a new cookie instead. In other words, all cookies are always valid and immutable state objects by default.

I'm not strongly opinionated on this, however - I just don't see any obvious use cases where this would both (1) not hold true and (2) simultaneously allow forking a session. For example, if I have a single-use cookie and make a request with it, that cookie value is invalidated and I probably can't fork off it, either.

I do think explicit session creation is a good idea, however.

proposals/session-track.md
```diff
@@ -0,0 +1,19 @@
...
+
+**Core Principle: Adapted Cookie Semantics**
+
+* **RFC 6265 as a Foundation:** We will use **RFC 6265 (HTTP Cookies)** as a functional starting point to determine necessary adaptations for MCP.  
+* **Payload Integration:** Session data will primarily be included as a dedicated field within the JSON-RPC payload. We will look to **SEP-1655** and **SEP-1685** for inspiration on structuring this metadata.
+
+**Key Requirements**
+
+* **Routing Compatibility & Size Constraints:** To allow load balancers to route requests based on session affinity, we may need to copy the session identifier/token into an HTTP header. This introduces strict **size constraints**; the session token must remain small enough to fit within standard header limits (typically \~4KB), prohibiting large state blobs from being stored directly in the token.  
+* **Security & Integrity:** Servers utilizing cookies must encrypt or sign the data to prevent client-side tampering. We will evaluate including a standard mechanism for signing/encryption directly in the spec.
```
LucaButBoring, line 13: I agree with this, but I believe we need to explicitly justify why we want to go beyond RFC 6265 and require this when most applications use unsigned cookies all the time without problems. Some good reasons I can think of:
1. Unlike a website, MCP host applications are server-agnostic, and an agent or host application has little reason to directly alter a server-provided cookie in the first place.
2. This is (at least in isolation) privacy-preserving; it's not possible for to add fingerprinting to a signed cookie - of course, tools themselves don't have this property, so it's a bit moot, but at least it doesn't make things worse.

proposals/session-track.md
lmaccherone: One of my interest is WebSocket transport which has the same problem as stdio in that it doesn't have headers. I too want to see everything necessary to maintain session in the JSON-RPC payload. I also use HTTP POST transport but not SSE HTTP GET. I believe that spec is still quite difficult to implement and while I hope this work fixes that, I see WebSockets as a much better alternative to streaming with SSE HTTP.

proposals/session-track.md
lmaccherone: I'll take a stronger position against this requirement. It's added complexity. At the very least, I believe this should be no more than "MAY" with guidance on how to do it if they include it but I believe this does more harm than good. As Luca said, headers are sent unencrypted all the time, but I'll also add that the information we're talking about adding there is not particularly sensitive.

lmaccherone: On the other hand, requiring cookies to have these options: `Secure`, `HttpOnly`, `SameSite=Strict|Lax` makes sense. Also having a narrow path/domain restrictions is good practice.

proposals/session-track.md
lmaccherone: Yes to `session/create` and `session/destroy`. I too don't see a use case for `session/fork`

proposals/session-track.md
evalstate: Principle has been agreed that payload content can be duplicated to headers if wanted. STDIO would need "cookie" data in JSON-RPC

evalstate: Content is required in JSON-RPC package for transports that don't have header-like side-channels (WebSockets, gRPC etc).

proposals/session-track.md
evalstate: Absolute minimum is SessionID/Version. Resumption/Last-Known-Id. <-- next layer of design. 

---

1) Limit to SessionId Only (minimum for supporting routing)
2) Keep the session envelope open with an arbitrary size limit. 
3) Keep it open with no size limit

Constrained by common infrastructure limitations. Question to infra specialists -> what are the hard constraints.

evalstate: What is the SDK experience from a Client/Server perspective.

proposals/session-track.md
evalstate: Do we need security beyond the current specification level (e.g. signing and encryption of content). This is not an application layer concern.

proposals/session-track.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Prior Art
+
+This section is for summarising the relevant parts of how other Agent systems and specifications handle Sessions.
+
+### Agent Client Protocol (Specification)
+
+Sessions are created by the Client (Host) and resumable, with the option of the Agent replaying messages to rehydrate Client state. There is no "close" operation. 
+
+Sessions are well specified with a unique session identifier, multiplexing and sequences of Prompt Turns between Client and Agent. Sessions in ACP are a user level abstraction. ACP is currently STDIO only with an HTTP transport "in progress".
```
evalstate, line 33: ss-> interesting design, not sure how related this specifically is to MCP. It is a high level application protocol.

proposals/session-track.md
```diff
@@ -0,0 +1,118 @@
...
+
+https://agentclientprotocol.com/rfds/session-list
+
+##### Session Forking
+
+https://agentclientprotocol.com/rfds/session-fork
+
+##### MCP/ACP transport (MCP over ACP)
+
+https://agentclientprotocol.com/rfds/mcp-over-acp
```
evalstate, line 47: Is reusing or incorporating an ACP Session Id worthwhile.

proposals/session-track.md
```diff
@@ -0,0 +1,118 @@
...
+
+Agents hosted as MCP Servers can have an instance type of `connection`, `request` or `shared`.
+
+### Tool Generates association identifier.
+
+A Tool Call returns a unique identifier, and requests that the identifier be returned in subsequent tool calls. 
+
+This pattern is potentially unreliable due to LLM generation being potentially unreliable but works in practice. 
+
+Promoting this technique to a determenistic side-channel is a possibility.
```
evalstate, line 100:
```suggestion
Promoting this technique to a deterministic side-channel is a possibility.
```

proposals/session-track.md
```diff
@@ -0,0 +1,118 @@
...
+
+**Key Requirements**
+
+* **Routing Compatibility & Size Constraints:** To allow load balancers to route requests based on session affinity, we may need to copy the session identifier/token into an HTTP header. This introduces strict **size constraints**; the session token must remain small enough to fit within standard header limits (typically \~4KB), prohibiting large state blobs from being stored directly in the token.  
+* **Security & Integrity:** Servers utilizing cookies must encrypt or sign the data to prevent client-side tampering. We will evaluate including a standard mechanism for signing/encryption directly in the spec.  
+* **Expert Review:** We will consult security experts to review the design for known vulnerabilities common in standard cookie implementations.
+
+**Key Design Decision: Lifecycle Management**
+
+> (Gabriel Z) Another key decision is around parallel tool calling. I think the decision was that parallel tool calling would just overwrite the session data nondeterministically if multiple tools decided to add session data. It could be good to at least add a warning to that effect.
```
markdroth, line 18: I think that there are two distinct use-cases here: per-session state, and per-tool-call state.

I think per-session state is in scope for the session track, and I think HTTP cookies can work pretty well for this.  The idea behind using cookies for session data is that the client can basically use a different cookie jar for each session, and it's totally up to the server as to how and when to update the cookies within that session.  So, for example, the session/create RPC would wind up creating a cookie indicating the session ID, which the client would then store in its cookie jar and therefore send with every subsequent RPC in that session.  The server can decide in its response to any of those RPCs that it wants to set an additional cookie, and that's fine -- there's a fair amount of flexibility there for server implementations.  (Clients that don't support sessions would just use a single cookie jar for all RPCs.)

I don't think per-tool-call state is in scope for the session track; instead, I think that should be addressed as part of the multi round-trip requests track, and there's already [some discussion about this in the multi round-trip request track's brief doc](https://docs.google.com/document/d/1i12c1NXhDBaCQYlL-GOpC2fh6Yr_2itlc07WP7-JqAE/edit?disco=AAABxrTqQE0).  To summarize, I don't think the HTTP cookie approach will work very well for per-tool-call state, specifically because there can be multiple tool calls in flight in parallel (either within a session or not).  The server implementation would need some way to know what cookie name to use for the state for each in-flight tool call, or else the cookie data would simply conflict, so we'd need to provide some way to do that.  And it also doesn't seem to scale very well to have every tool call include cookies for state for every other tool call that is in flight at the same time.  I think that instead, we'll wind up wanting something more like SEP-1685 here.

PR #7, opened 2026-01-23
Multi Round-Trip Request (MRTR) Proposal

markdroth:
Initial draft of a proposal for the MRTR problem.

See previous discussion in the [original document](https://docs.google.com/document/d/1EQSttkKGOUNXzJM998VDhBu3288f4IqRM0jd1tWbdos/edit?usp=sharing).

### DIFF
--- a/proposals/multi-round-trip-requests-proposal.md
+++ b/proposals/multi-round-trip-requests-proposal.md
@@ -1,42 +1,55 @@
 # MCP Streamless Round-Trip Proposal Sketch
 
-Author: [Mark Roth](mailto:roth@google.com)  
-Last modified: 2025-12-02
+## Background
 
-# Background
+For context, see [MCP Multi Round-Trip Requests Track
+Brief](https://github.com/modelcontextprotocol/transports-wg/pull/5).
 
-For context, see [MCP Multi Round-Trip Requests Problem Statement](https://docs.google.com/document/d/1trA6-b8nrhE-yZQ2BPNltCSlxhkAmVQsG9a6pPzH6Fs/edit?usp=sharing).
+This document describes an approach to that problem that does not require
+the use of SSE streams.  This makes it possible to use server-initiated
+requests like elicitations without needing to support SSE streams.
 
-This document describes an approach to that problem that does not require the use of SSE streams.  This makes it possible to use server-initiated requests like elicitations without needing to support SSE streams.
-
-# Proposal
+## Proposal
 
 There is a common MCP use case that looks like this:
 
-1. Client issues a tool call to an MCP server.  
-2. In order to respond to that request, the server needs more info from the client, so it sends back a request for more info (e.g., a sampling or elicitation request).  
-3. The client responds to the server's request for more info.  
+1. Client issues a tool call to an MCP server.
+2. In order to respond to that request, the server needs more info from the
+   client, so it sends back a request for more info (e.g., a sampling or
+   elicitation request).
+3. The client responds to the server's request for more info.
 4. The server can now respond to the original request from the client.
 
 Today in MCP, this is handled as follows:
 
-1. Client sends an HTTP request.  
-2. Server responds to the HTTP request from step 1 by opening an SSE stream.  On that SSE stream, it sends the request for more info back to the client.  
-3. Client sends a separate, independent HTTP request with the response to the server's request from step 2\.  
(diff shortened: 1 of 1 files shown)


proposals/multi-round-trip-requests-proposal.md
```diff
@@ -0,0 +1,210 @@
...
+to handle steps 2 and 3.
+
+In the TS schema, this might look something like the following (not
+quite completely right, but provides the general idea):
+
+```ts
+// Similar to existing JSONRPCResultResponse.
+// Used in cases where the server needs the results of one or more requests
+// of its own before it can complete the client's request.
+export interface JSONRPCIncompleteResultResponse {
```
CaitieM20, line 58: Agree with the proposal for tool calls,

Would like to discuss the Result Response. In https://github.com/modelcontextprotocol/transports-wg/pull/9 I propose two additional message format options, that extend or reuse existing concepts vs adding a new one.

My preference is Option 2 which extends the existing [Tool Result Concept](https://modelcontextprotocol.io/specification/2025-03-26/server/tools#tool-result) to encapsulate this new response type while being able to leverage existing messaging semantics.

proposals/multi-round-trip-requests-proposal.md
```diff
@@ -0,0 +1,210 @@
...
+   the server's request from step 2.
+4. Server sends the response to the client's original request from step 1
+   on the SSE stream that it opened in step 2.
+
+This document proposes the following new approach for this use-case:
+
+1. Client sends a CallToolRequest.
+2. Server sends back a single response (**not** an SSE stream) indicating
+   that the request is incomplete and that an elicitation request is required.
+3. Client sends a new CallToolRequest, completely independent of the
```
gjz22, line 38: Edge case to consider: The server sends multiple requests. The client responds with one. Some options:
1. [Preferred] Return a `400`. The client is required by protocol to return all responses at the same time.
2. The client responds with the requests not answered. This seems less ideal as it's possible a client could answer the same server request multiple times.

proposals/multi-round-trip-requests-proposal.md
```diff
@@ -0,0 +1,210 @@
...
+```ts
+// Similar to existing JSONRPCResultResponse.
+// Used in cases where the server needs the results of one or more requests
+// of its own before it can complete the client's request.
+export interface JSONRPCIncompleteResultResponse {
+  jsonrpc: typeof JSONRPC_VERSION;
+  id: RequestId;
+  // Requests issued by the server that must be complete before the
+  // client can retry.
+  dependent_requests: { [key: string]: ServerRequest };
```
gjz22, line 63: Proposal rather than a key:
Let's use 1685 for this. The goal of 1685 is to allow the server to pass context to the client that is echo'ed back to the server when the response is submitted in order to help the server process the response. That's effectively what is being done here. If the state is passed to the client, they then CAN use a machine readable key but they don't have to in cases where there is only one request (likely the vast majority), which simplifies clients and reduces complexity and allows us to keep this as a list.

Example: `state: "github_login"`

Separately, a ServerRequest is a JSON RPC request, so it will seem weird to nest them. For example, should the JSON RPC requirements that an id is supplied in the request and the response should mirror that be respected? Should we add a new type here, that is more clear/useful long term?

Randgalt: +1 not having a unique ID for the dependent requests will be hard to implement in servers

proposals/multi-round-trip-requests-proposal.md
Randgalt: +1 to extending Tool Results. It's already polymorphic so adding new "request" types would be easier to do while maintaining backward compatibility when writing servers.

markdroth:
Closing this as a duplicate of #12.

proposals/multi-round-trip-requests-proposal.md
markdroth: The problem is that I don't think the server can tell whether the client responded to all of the requests it was told to, because it doesn't know which server instance told it to send those responses.  In upgrade cases, this may be difficult or impossible to tell.

I think it's fine for the server to just respond with whichever dependent requests it doesn't yet have.  If the client sees the same request a second time because it didn't answer it the first time, that just means that the client can deal with the result of its own bug. :)

markdroth (commented):
Just responding to one quick comment here that I didn't notice earlier.  After this, let's take discussion to #12.  Thanks!

PR #9, opened 2026-01-27
Message Structure Proposal to align MRTR & Tasks

CaitieM20:
Updated pros and cons for MRTR & Tasks message structure options, This is an addendum to [Multi Round-Trip Request (MRTR) Proposal](https://github.com/modelcontextprotocol/transports-wg/pull/7).

This proposal is aligned with MRTR proposal and dives deeper into options on message structure and ensuring the changes align with Tasks.

It is informed by a test implementation of an EchoTool that can work as a standard Tool, Task and also request Elicitations.

### DIFF
--- a/proposals/MRTR-And-Tasks-Message-Structure-Proposal.md
+++ b/proposals/MRTR-And-Tasks-Message-Structure-Proposal.md
@@ -251,8 +251,8 @@ This could look like, and would replace the response in step 6 above:
 Pros: 
 - Consistent message structure for all Result types from Tasks & Tools which simplifies the SDK implementations. 
 - Consistent handling of isError or other future result metadata fields.
-- With Tasks this structure would allow for partial results & elicitation & sampling requests to be part of the same message.
-- Supports multiple Elicitation & Sampling requests at the same time.  Notably with Tasks this structure would allow for partial results to also be retured as part of the 'result' message in addition to Elicitation or Sampling requests. 
+- With Tasks this structure would allow for partial results & additional input to requested on the same get/results 
+- Supports multiple Elicitation & Sampling requests at the same time.  
 
 Cons: Larger change to existing implementations. 
 - Requires deprecating the existing `Elicitation` and `Sampling` messages returned by Tasks since out of band messages are no longer needed. 
--- a/proposals/MRTR-And-Tasks-Message-Structure-Proposal.md
+++ b/proposals/MRTR-And-Tasks-Message-Structure-Proposal.md
@@ -5,16 +5,16 @@ This document provides an overview of how the [Multi-Round-Trip Request Proposal
 
 A few notes the Goals of MRTR are to:
 1. Ensure the protocol itself is stateless while allowing for stateful application semantics
-2. Eliminating the need for the SSE stream, this is marked as Optional in the transport today so should only be used for optimization/performance not required.
+2. Making the GET SSE stream truly optional, i.e. not required for core functionality
 3. Remove the ability for sampling/elicitation to be sent out of band (i.e. without a client request).
 
 In Tasks this means:
 1. TaskId is used to represent the application state
-2. Remove the guidance around using SSE stream, and model the call flow as request/response semantics.
+2. Remove the guidance around using SSE stream, and model the call flow as request/response semantics only.
 
 Tasks & Tool Calls provide mecahnisms for implementing two different kinds of messaging patterns. 
-1. Tool Calls: short running/sync - i.e. return within ms/seconds, low cost to compute/answer
-2. Tasks: long running/async - i.e. can run for minutes to hours, higher cost to compute.
+1. Tool Calls: short running/sync/stateless - i.e. return within ms/seconds, low cost to compute/answer. 
+2. Tasks: long running/async/stateful - i.e. can run for minutes to hours, higher cost to compute. Server may be storing state between calls related to TaskId.
 
 Given the above these patterns will manifest MRTR in two different ways.
 1. Tool Calls: Server sends a Result with an elicitation/sampling request. The server stops processing the request at this point, and the client must retry
@@ -85,7 +85,7 @@ The below example uses an Echo Tool with an optional input parameter, when missi
     "result":  
     {
         "taskId":  "echo_dc792e24-01b5-4c0a-abcb-0559848ca3c5",
-        "status":  "input_require",
+        "status":  "input_required",
         "statusMessage":  "Input Required to Proceed call tasks/result",
         "createdAt":  "2026-01-27T03:38:07.7534643Z",
         "ttl":  60000,
--- a/proposals/MRTR-And-Tasks-Message-Structure-Proposal.md
+++ b/proposals/MRTR-And-Tasks-Message-Structure-Proposal.md
@@ -1,0 +1,259 @@
+# MRTR And Tasks Message Structure Proposal
+
+## Overview
+This document provides an overview of how the [Multi-Round-Trip Request Proposal](https://github.com/modelcontextprotocol/transports-wg/pull/7/changes#diff-c42674696a4c91ccc0d2daf8425dbcb52201ec1ef75921ae1e4865b5b911018d) (MRTR) fits with the Tasks by walking thorugh an example with Elicitations.
+
+It also raises open questions and proposed solutions for discussion on what the return type should be for Tool Requests that require Elicitation or Sampling to complete.
+
+## Tasks Background
+The [Tasks](https://modelcontextprotocol.io/specification/draft/basic/utilities/tasks) Utilities allows requests to be augmented with a Promise like mechanism. `Tasks` have a Status Lifecycle including `Working`, `Input Requried`, `Completed`, `Failed`, and `Cancelled`.
+
+The `Input Required` status allows a Tool or Capability to indicate that additional input is required from the user to complete the task. This is where an Elicitation or Sampling request can be made by the server.
+
+When the client encounters an `Input Required` status it SHOULD call `tasks/result`. This allows the server to then return an `Elicitation` or `Sampling` request to the client. This fits well with the proposal to eliminate unsolicited `Elicitation` and `Sampling` requests from the server to the client.
+
+### Example Flow with Elicitations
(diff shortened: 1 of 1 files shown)


proposals/MRTR-And-Tasks-Message-Structure-Proposal.md
```diff
@@ -0,0 +1,259 @@
...
+    "params": {
+        "taskId": "echo_dc792e24-01b5-4c0a-abcb-0559848ca3c5"
+    }
+}
+```
+
+6. <b>Server Response</b> returns `elicitation/create` to request additional input
+```json
+{
+    "id":  3,
```
LucaButBoring, line 112: This isn't quite right as stated - this is:
- a server-to-client request (with its own request ID)
- that occurs as a side-effect of a `tasks/result` client-to-server request

In Tasks, we're not altering the fundamental request/response flow (I mean we are actually with `CreateTaskResult`, but that's it) - what we're doing is creating an opportunity for the server to open an SSE stream to _side-channel_ an elicitation request on, independent of the final task result itself. The server is actually allowed to open this side-channel in response to any request (or even send messages on the background stream), but we specifically require `tasks/result` to block until a terminal state to reserve it for this purpose.

The sequence diagram from the spec shows the client terminating the stream after receiving the elicitation request, but that's not actually required even though it is allowed.

proposals/MRTR-And-Tasks-Message-Structure-Proposal.md
```diff
@@ -0,0 +1,259 @@
...
+        }
+    } 
+}
+```
+
+7. <b>Client Request</b> presents the Elicitation to the user and collects the input, then sends message to the server.
+```json
+{
+    "jsonrpc": "2.0",
+    "id": 4,
```
LucaButBoring, line 142: Should reflect the request ID of the elicitation request

proposals/MRTR-And-Tasks-Message-Structure-Proposal.md
```diff
@@ -0,0 +1,259 @@
...
+        "pollInterval":  100
+    },
+}
+```
+
+11. <b>Client Request</b> calls `tasks/result` to get the final result of the `Task` from the server.
+Client Message
+```json
+{
+    "id":  6,
```
LucaButBoring, line 192: The client is also allowed to keep a stream from the initial `tasks/result` request open, in which case the server response will have an `id` of 3 (as that would be the response to the `tasks/result` request).

proposals/MRTR-And-Tasks-Message-Structure-Proposal.md
```diff
@@ -0,0 +1,259 @@
...
+                    {
+                        "input": {"type":  "string"}
+                    },
+                    "required": ["input"]
+                }
+            }],
+        "isError": false,
+    }
+}
+```
```
LucaButBoring, line 266: How would the client send the response back to the server, here?

proposals/MRTR-And-Tasks-Message-Structure-Proposal.md
```diff
@@ -0,0 +1,259 @@
...
+4. <b>Server Response</b> with Task status `InputRequired`
+```json
+{
+    "id":  2,
+    "jsonrpc":  "2.0",
+    "result":  
+    {
+        "taskId":  "echo_dc792e24-01b5-4c0a-abcb-0559848ca3c5",
+        "status":  "InputRequired",
+        "statusMessage":  "Input Required to Proceed call get/result",
```
halter73: Nit:


```suggestion
        "status":  "input_require",
        "statusMessage":  "Input Required to Proceed call tasks/result",
```

proposals/MRTR-And-Tasks-Message-Structure-Proposal.md
halter73: > The sequence diagram from the spec shows the client terminating the stream after receiving the elicitation request, but that's not actually required even though it is allowed.

Does the TS SDK client terminate the result stream after receiving a request? I assume not. If it did, it would by heavily relying on `Last-Event-Id` based stream resumability to work so it doesn't miss any messages when it comes back, and I'm not sure how widely supported that is even for MCP servers that support tasks.

proposals/MRTR-And-Tasks-Message-Structure-Proposal.md
halter73: 👍 That probably means that the `elicitation/create` request should get a new ID rather than reusing the `tasks/result` request ID.

proposals/MRTR-And-Tasks-Message-Structure-Proposal.md
halter73: I assume as part of the `"dependent_responses"` as described in #7 for MRTR or continue as part of an independent request for Tasks.

proposals/MRTR-And-Tasks-Message-Structure-Proposal.md
CaitieM20: @LucaButBoring ack on how it behaves today. One of the goals with the MRTR is to eliminate Server initiated messages, so this is the proposal on how the request flow would change to accommodate that today. 

Also in the December meetup we discussed having the SSE be an option for performance but not a required part of functionality since it is Optional in the protocol itself. Additionally the current behavior adds a lot of complexity for what could be a request/response pattern.

proposals/MRTR-And-Tasks-Message-Structure-Proposal.md
CaitieM20: Gotcha in how it works today, see above I'm trying to model how this would work without requiring the SSE stream since its marked as Optional in the spec.

proposals/MRTR-And-Tasks-Message-Structure-Proposal.md
CaitieM20: Good call outs, I think this is a place where the spec is ambiguous and we should align ourselves here.  Right now we have MCP-Session-Id (slated to be removed), MessageId, and TaskId all encapsulating some state. 

Is the JsonRPC Envelope Id a piece of state that persists across sessions, or is it something that is tied to request/response flow.

proposals/MRTR-And-Tasks-Message-Structure-Proposal.md
CaitieM20: My understanding is that Elicitation responses get sent to the POST /mcp endpoint today anyway. It would include the task metadata to be routed to the right task on the server handling the request. Because we have the task_id as the unifying session/state identifier for this request we can use that to route, and don't need a persistent connection, MCP-Session-Id or the Json-RPC Id. 

This is how I have it in my prototype implementation.

proposals/MRTR-And-Tasks-Message-Structure-Proposal.md
LucaButBoring: > > The sequence diagram from the spec shows the client terminating the stream after receiving the elicitation request, but that's not actually required even though it is allowed.
> 
> Does the TS SDK client terminate the result stream after receiving a request? I assume not. If it did, it would by heavily relying on `Last-Event-Id` based stream resumability to work so it doesn't miss any messages when it comes back, and I'm not sure how widely supported that is even for MCP servers that support tasks.

The client does not proactively terminate streams, correct - that flow can use regular stream resumability there, too (though with MRTR we might be able to do away with even that for the most-common use cases).

proposals/MRTR-And-Tasks-Message-Structure-Proposal.md
LucaButBoring: > Is the JsonRPC Envelope Id a piece of state that persists across sessions, or is it something that is tied to request/response flow.

From [requests](https://modelcontextprotocol.io/specification/2025-06-18/basic#requests):

> The request ID **MUST NOT** have been previously used by the requestor within the same session.

But notably, this part of the spec doesn't actually define what a session is. I am at least certain that multiple clients can use the same request IDs as each other with a given server.

proposals/MRTR-And-Tasks-Message-Structure-Proposal.md
LucaButBoring: I'm more wondering how separate elicitation requests are kept unique - today that's via request IDs, and in the MRTR example in #7, that's via named identifiers, but there doesn't seem to be anything equivalent here. A task can have multiple elicitations, so the task ID alone isn't enough to disambiguate them.

proposals/MRTR-And-Tasks-Message-Structure-Proposal.md
```diff
@@ -0,0 +1,276 @@
+# MRTR And Tasks Message Structure Proposal
+
+## Overview
+This document provides an overview of how the [Multi-Round-Trip Request Proposal](https://github.com/modelcontextprotocol/transports-wg/pull/7/changes#diff-c42674696a4c91ccc0d2daf8425dbcb52201ec1ef75921ae1e4865b5b911018d) (MRTR) fits with the Tasks by walking thorugh an example with Elicitations.
```
evalstate:
```suggestion
This document provides an overview of how the [Multi-Round-Trip Request Proposal](https://github.com/modelcontextprotocol/transports-wg/pull/7/changes#diff-c42674696a4c91ccc0d2daf8425dbcb52201ec1ef75921ae1e4865b5b911018d) (MRTR) fits with the Tasks by walking through an example with Elicitations.
```

proposals/MRTR-And-Tasks-Message-Structure-Proposal.md
```diff
@@ -0,0 +1,276 @@
...
+
+4. <b>Server Response</b> with Task status `InputRequired`
+```json
+{
+    "id":  2,
+    "jsonrpc":  "2.0",
+    "result":  
+    {
+        "taskId":  "echo_dc792e24-01b5-4c0a-abcb-0559848ca3c5",
+        "status":  "input_require",
```
evalstate:
```suggestion
        "status":  "input_required",
```

CaitieM20:
Closing this as the discussion and proposal is incorporated into [SEP-XXXX: Multi Round-Trip Requests](https://github.com/modelcontextprotocol/transports-wg/pull/12)

PR #12, opened 2026-02-03
SEP-XXXX: Multi Round-Trip Requests

markdroth:
This is a draft SEP for Multi Round-Trip Requests (MRTR).

## Motivation and Context
See SEP for details.

## How Has This Been Tested?
Not tested yet, that will happen when we implement.

## Breaking Changes
See SEP for details.

## Types of changes
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

## Additional context
N/A


proposals/XXXX-MRTR.md
```diff
@@ -0,0 +1,707 @@
...
+The workflow for `Tasks` is as follows:
+
+1. Server sets Task Status to `input_required` 
+2. Client retrieves the Task Status by calling `tasks/get` and sees that more information is needed.
+3. Client calls `task/result` 
+4. Server returns the `DependentRequets` object. The Server can pause processing the request at this point.
+5. Client sends `DependentResponses` object to server along with `Task` metadata field.
+6. Server resumes processing sets TaskStatus back to `Working`.
+
+Since `Tasks` are likely longer running, have state associated with them, and are likely more costly to compute, the request for more information does not end the original request. Instead, the server can resume processing once the necessary information is provided.
```
clareliguori: Nit on `does not end the original request`: I'm getting confused between a client http request like `tasks/get` vs what I think you are referring to here, the overall `Task` request.

markdroth: I've attempted to clarify the wording here.

proposals/XXXX-MRTR.md
```diff
@@ -0,0 +1,995 @@
...
+those requests.  The individual requests and responses are stored in a
+map with string keys:
+
+```typescript
+export interface InputRequests { [key: string]: ServerRequest; }
+
+export interface InputResponses { [key: string]: ClientResult; }
+```
+
+TODO: The above schema definitions are not quite right, because
```
CaitieM20, line 172: This TODO is just for the SDK right? Example is TypeScript and likely indicates refactoring needed by other SDKs right?

proposals/XXXX-MRTR.md
```diff
@@ -0,0 +1,995 @@
...
+context of a client request.  This new mechanism will have a slightly
+different workflow for ephemeral tools and persistent tools, the latter
+of which will leverage Tasks.  However, both workflows will use the same
+data structures.
+
+First, we introduce the notion of "input requests", which represents
+a set of one or more server-initiated request to be sent to the client,
+and "input responses", which represents the client's responses to
+those requests.  The individual requests and responses are stored in a
+map with string keys:
```
CaitieM20:
```suggestion
map with string keys. The values of the map are protocol requests/responses. Examples include Elicitation & Sampling. An example of how this would look in the TypeScript SDK is below:
```
recommend calling out the map values.  I think you can infer it from below but always good to be specific.

proposals/XXXX-MRTR.md
```diff
@@ -0,0 +1,995 @@
...
+   This terminates the original request.
+3. Client sends a new tool call request, completely independent of the
+   original one.  This new tool call includes responses to the input
+   requests from step 2.  It also includes the request state specified by
+   the server in step 2.
+4. Server sends back a CallToolResponse.
+
+Note that the requests in steps 1 and 3 are completely independent: the
+server that processes the request in step 3 does not need any
+information that is not directly present in the request.
```
CaitieM20:
```suggestion
information that is not directly present in the request. To support this decoupling the JsonRPC Id MUST be different between the requests sent in step 1 and step 3. 
```

proposals/XXXX-MRTR.md
markdroth: Done.

proposals/XXXX-MRTR.md
markdroth: No, unfortunately, it's the actual schema definition, not the SDK.

As an example, in the current schema, [`ServerRequest`](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/99a2ff0b218dc46834ccd3a37a0df1960118db8e/schema/2025-11-25/schema.ts#L2549) is basically a list of all possible server-initiated request types, such as [`ElicitRequest`](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/99a2ff0b218dc46834ccd3a37a0df1960118db8e/schema/2025-11-25/schema.ts#L2229), which inherits from [`JSONRPCRequest`](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/99a2ff0b218dc46834ccd3a37a0df1960118db8e/schema/2025-11-25/schema.ts#L131), which includes the id field.

To deal with this, we'll need to do a bit of refactoring here, probably something like this:
- Define a new `JSONRPCInputRequest` base type that is the same as `JSONRPCRequest` but without the request ID.
- Have a new variant of `ElicitRequest` called `ElicitInputRequest` that inherits from `JSONRPCInputRequest` instead of from `JSONRPCRequest`.  (Same for sampling, etc.)
- Define a new `ServerInputRequest` type that is the same as `ServerRequest`, except that it uses the new `*InputRequest` types instead of the original types.
- Define `InputRequests` to use the new `ServerInputRequest` type instead of `ServerRequest`.

I think this should be straightforward, but including the details here seemed like they would distract from the real intent of this document more than they'd be useful, so I was planning to defer these details until we vote this out of the transport WG and I put together the real SEP in the main repo.

Happy to revisit this if we consider it a problem.

proposals/XXXX-MRTR.md
```diff
@@ -0,0 +1,998 @@
...
+def my_tool(request):
+  github_login = request.input_responses().get('github_login', None)
+  if github_login is None:
+    return IncompleteResponse({'github_login': elicitation_request})
+  result = GetResult(github_login)
+  return Result(result)
+```
+
+However, we need to consider how to avoid breaking things for existing
+tools that are written the old way.  Ideally, we will be able to modify
```
pja-ant, line 1165: Hmm, this "ideally" is doing a lot of heavy lifting. I don't think we can accept this without a full story here? It seems really non-trivial to make this just work, and if we can't make it just work then this becomes a breaking change for lots of people, which could be a major problem.

I think before we move forward with this SEP we need to figure out what's happening here.

markdroth: I agree that we need to figure this out, and I think we'll need help from some of the SDK maintainers to do so.  I suspect that there won't be a single answer here; I think we'll need a story in each language, and they may not all be the same.  If you can help get the right set of people involved, I'm happy to provide input.

That having been said, I'm not sure that this issue needs to block the SEP from moving forward.  One question I would ask here is, if there is some subset of languages in which this change may be a bit disruptive, are we going to reject this SEP as a result?  I'm not sure we can afford to do that -- this change is badly needed, and I don't know that we have a good alternative.  And if not, it seems like we shouldn't prevent this from moving forward while we investigate this.

We can probably discuss this further at today's meeting.

proposals/XXXX-MRTR.md
```diff
@@ -0,0 +1,998 @@
...
+the SDKs to support the old tool implementations via some sort of
+backward compatibility layer.
+
+## Security Implications
+
+Because `request_state` passes through the client, malicious or
+compromised clients could attempt to modify it to alter server behavior,
+bypass authorization checks, or corrupt server logic.  To mitigate this,
+we require servers to validate this state somehow.
```
pja-ant: yeah probably need to require some sort of signing or something - I think the original state SEP had recommendations for this that looked good.

proposals/XXXX-MRTR.md
```diff
@@ -0,0 +1,998 @@
...
+
+These types will be used in two different workflows, one for ephemeral
+tools and another for persistent tools.
+
+### Ephemeral Tool Workflow
+
+For ephemeral tools, we will adopt the following workflow:
+
+1. Client sends tool call request.
+2. Server sends back a single response (**not** an SSE stream)
```
pja-ant: Can it optionally be an SSE stream so that we can use progress? i.e. leave it up to the server to decide what to respond with.

pja-ant: If the server doesn't like long-lived connections then they can just choose to return a single response.

CaitieM20: So for ephemeral tools we are saying there is no more progress being made on the server for this particular execution to make this stateless. Without ending processing on the initial request you still have the problem of how do I associate the Elicitation/Sampling Results to the server processing the initial request.

proposals/XXXX-MRTR.md
CaitieM20: not a problem, we can follow up once on tackling schema changes once we get approval on general direction.

proposals/XXXX-MRTR.md
```diff
@@ -0,0 +1,998 @@
...
+   - Examples: weather app, accessing email
+2. **Persistent**: State is accumulated on the server side.
+   - Server may generate a large amount of state before requesting more
+     info from the client, and it may need to pick up that state to
+     continue processing after it receives the info from the client.
+   - Server may need to continue processing in the background while
+     waiting for more info from the client, in which case server-side
+     state is needed to track that ongoing processing.
+   - Examples: accessing an agent, spinning up a VM and needing user
+     interaction to manipulate the VM
```
mikekistler, line 46: Just stated above is that tools are simply use for exposition here, and the SEP is not specific to tool calls. Are the "two types" described here common across client-to-server requests, or specific to tool calls? If specific to tool calls, are all other client-to-server requests one or the other of these, or a different flavor altogehter?

proposals/XXXX-MRTR.md
CaitieM20: I suspect this will require some breaking change to get to a stateless protocol, but agree we need a clear migration path here.

proposals/XXXX-MRTR.md
```diff
@@ -0,0 +1,998 @@
...
+  jsonrpc: typeof JSONRPC_VERSION;
+  id: RequestId;
+  // Requests issued by the server that must be complete before the
+  // client can retry the original request.
+  input_requests?: InputRequests;
+  // Request state to be passed back to the server when the client
+  // retries the original request.
+  // Note: The client must treat this as an opaque blob; it must not
+  // interpret it in any way.
+  request_state?: string;
```
pja-ant: nit: these should be `lowerCamelCase` rather than `snake_case`

proposals/XXXX-MRTR.md
```diff
@@ -0,0 +1,998 @@
...
+common for tools to be deployed in a horizontally scaled, load balanced
+service, so we need to optimize for this case.
+
+Today, if a tool needs to send an elicitation request in order to make
+progress, the workflow works like this:
+
+1. Client sends tool call request.  For this example, let's assume that
+   the load balancers happen to send this request to server instance A.
+2. Server A opens an SSE stream and sends the elicitation response on that
+   stream.
```
mikekistler, line 58: It is also possible that the elicitation request is sent in the GET stream if it is available, but maybe that's not worth mentioning.

proposals/XXXX-MRTR.md
```diff
@@ -0,0 +1,998 @@
...
+     `input_requests` field.
+   - The `JSONRPCIncompleteResultResponse` message MAY include a
+     `request_state` field.  If specified, this field is an opaque
+     string that is meaningful only to the server.  Servers are free to
+     encode the state in any format (e.g., plain JSON, base64-encoded
+     JSON, encrypted JWT, serialized binary, etc.).
+   - If a request contains a `request_state` field, servers MUST always
+     validate that state, as the client is an untrusted intermediary.
+     If tampering is a concern, servers SHOULD encrypt the `request_state`
+     field (e.g., using AES-GCM or a signed JWT) to ensure both
```
kurtisvg: Is it up to the server on how to encrypt it (these are just recommendations)?

proposals/XXXX-MRTR.md
```diff
@@ -0,0 +1,998 @@
...
+### Persistent Tool Workflow
+
+The persistent tool workflow will leverage Tasks. [`Tasks`](https://modelcontextprotocol.io/specification/draft/basic/utilities/tasks) already provide a mechanism to indicate that more information is needed to complete the request. The `input_required` Task Status allows the server to indicate that additional information is needed to complete processing the task. 
+
+The workflow for `Tasks` is as follows:
+
+1. Server sets Task Status to `input_required` 
+2. Client retrieves the Task Status by calling `tasks/get` and sees that more information is needed.
+3. Client calls `task/result` 
+4. Server returns the `InputRequets` object. The Server can pause processing the request at this point.
```
kziemski: typo. InputRequests

kurtisvg: Wouldn't the server pause the request at the input required stage?

proposals/XXXX-MRTR.md
markdroth: I think this hinges on the open discussion about how much support for SSE streams we want to keep.  We talked about this in the context of #11, and we don't currently seem to have consensus on whether to remove progress notifications on the SSE stream or to support both that mechanism and the tasks-based one.  I am personally in favor of removing the old mechanism so that we can reduce our reliance on SSE streams and not require clients to support two ways of doing the same thing.  However, we need more discussion on that to come to consensus.

There's nothing in this SEP specifically that would require not sending this on an SSE stream, and if we want to kick this can down the road a bit, I'm open to changing this to say that it can be sent on an SSE stream.  But if it is, then it would be the final request on the stream, and the stream would be closed immediately after sending this response.

proposals/XXXX-MRTR.md
```diff
@@ -0,0 +1,998 @@
...
+5. Client sends `InputResponses` object to server along with `Task` metadata field.
```
kurtisvg: is this part of the task/result call?

proposals/XXXX-MRTR.md
markdroth: Yeah, there are some specific recommendations for this earlier in the doc, in the "protocol requirements" section.

proposals/XXXX-MRTR.md
```diff
@@ -0,0 +1,998 @@
...
+```
+
+Note that both the "input_requests" and "request_state" fields affect
+only the client's next retry of the original request.  They will not
+be used for any other request that the client may be sending in parallel
+(e.g., a tool list or even another tool call).
+
+Note that this workflow eliminates the need for the
+`URLElicitationRequiredError` error code.  That code will be removed
+from the spec.
```
mikekistler, line 328: How does URL Elicitation work in this new flow?

proposals/XXXX-MRTR.md
```diff
@@ -0,0 +1,998 @@
...
+1. Server sets Task Status to `input_required` 
+2. Client retrieves the Task Status by calling `tasks/get` and sees that more information is needed.
+3. Client calls `task/result` 
+4. Server returns the `InputRequets` object. The Server can pause processing the request at this point.
+5. Client sends `InputResponses` object to server along with `Task` metadata field.
+6. Server resumes processing sets TaskStatus back to `Working`.
+
+Since `Tasks` are likely longer running, have state associated with them, and are likely more costly to compute, the request for more information does not end the originally requested operation (e.g., the tool call). Instead, the server can resume processing once the necessary information is provided.
+
+To align with MRTR semantics, the server will respond to the `task/result` request with a `InputRequests` object. Both of these will have the same JsonRPC `id`. When the client responds with a `InputResponses` object this is a new client request wit a new JSONRPC `id` and therefore needs a new method name. We propose `tasks/input_response`.
```
LucaButBoring, line 750: Rather than having the client make a request to `tasks/result` for the server requests, and then respond on a separate `tasks/input_response` method, I think we should consider consolidating that logic into a single method. The `input_required` to `tasks/result` transition has always been a weird workaround for SSE only and I think we should get rid of that if possible.

proposals/XXXX-MRTR.md
pja-ant: ah yeah, just saw it 👍

proposals/XXXX-MRTR.md
pja-ant: I don't believe that's correct. Elicitations in response to a tool call (or other client request) have always needed to come on the POST SSE stream. GET is only for unsolicited (which we're removing)

proposals/XXXX-MRTR.md
```diff
@@ -0,0 +1,998 @@
...
+```json
+{
+    "jsonrpc": "2.0",
+    "id": 1,
+    "method": "tools/call",
+    "params":{
+        "name": "echo",
+        "task":{
+            "ttl": 60000
+        }
```
kurtisvg, line 767: Does this imply that if an input is a task or not is decided by the client?

proposals/XXXX-MRTR.md
```diff
@@ -0,0 +1,998 @@
...
+      "status": "input_required",
+      "statusMessage": "Input Required to Proceed call tasks/result",
+      "createdAt": "2026-01-27T03:38:07.7534643Z",
+      "ttl": 60000,
+      "pollInterval": 100
+    },
+}
+```
+
+5. <b>Client Request</b> sends message `tasks/result` to discover what input is required to proceed.
```
kurtisvg, line 818: Why did we land on using tasks/result here instead of specifying the input required in the task description?

proposals/XXXX-MRTR.md
pja-ant: FWIW, for session ID, we use "SHOULD" language:

> The session ID SHOULD be globally unique and cryptographically secure (e.g., a securely generated UUID, a JWT, or a cryptographic hash).The session ID SHOULD be globally unique and cryptographically secure (e.g., a securely generated UUID, a JWT, or a cryptographic hash).

proposals/XXXX-MRTR.md
markdroth: I've updated the wording to reflect that sending the incomplete response on an SSE stream is still allowed but not preferred.

proposals/XXXX-MRTR.md
markdroth: As we discussed, this would be handled exactly the same way as form elicitation requests: either way, the same elicitation request and elicitation response that would be used today are handled in the input requests and responses.

proposals/XXXX-MRTR.md
markdroth: Are you suggesting that the client use the `tasks/result` method to provide the input responses?  That seems a little odd to me, but I see the logic.  It's okay with me if that's our consensus.

@CaitieM20, thoughts?

proposals/XXXX-MRTR.md
markdroth: Currently, tasks require the client to opt in.  I think we agree that we want this to be decided by the server, and @LucaButBoring is proposing exactly that change in https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2229.  But since that's not part of this SEP, here we're sticking with the current approach.

proposals/XXXX-MRTR.md
markdroth: I don't think there's anything specific to tools here; I think it can all be generalized to any client-initiated request.  But it's easier to explain this doc if we talk about it specifically in terms of tool calls.

I've attempted to clarify the language above to make the intent clearer.

proposals/XXXX-MRTR.md
markdroth: Fixed.

proposals/XXXX-MRTR.md
markdroth: Clarified wording to indicate that the server can choose whatever encryption algorithm it wants.

proposals/XXXX-MRTR.md
markdroth: Fixed.

markdroth: Fixed.

proposals/XXXX-MRTR.md
markdroth: It's the new `tasks/input_response` RPC.  I've clarified the wording here.

proposals/XXXX-MRTR.md
markdroth: I've added a paragraph about this in the rationale section below.

markdroth (commented):
I've addressed the following points that we discussed in today's meeting:
  - allow progress on SSE stream followed by incomplete response
  - generalize wording to be non-tool-specific in motivation section
  - clarify wording on server encryption algorithms (it should be able to use any it wants)
  - move "client may stop processing at this point" from step 4 to step 1 of persistent workflow
  - note method name in step 5 for persistent
  - document why we didn't include the elicitation on tasks/get
  - highlight incomplete response with request_state but without input_requests

There is one other item that we talked about today but didn't come to consensus on, which is whether to support returning an error from the `tasks/provide_input` RPC.  That will need further discussion.

Please let me know if I missed anything or got anything wrong.  Thanks!

proposals/XXXX-MRTR.md
```diff
@@ -0,0 +1,1041 @@
...
+tools that are written the old way.  Ideally, we will be able to modify
+the SDKs to support the old tool implementations via some sort of
+backward compatibility layer.
+
+## Security Implications
+
+Because `requestState` passes through the client, malicious or
+compromised clients could attempt to modify it to alter server behavior,
+bypass authorization checks, or corrupt server logic.  To mitigate this,
+we require servers to validate this state somehow.
```
ramraaj25: I have a concern regarding the security of request_state in the stateless workflow.

Currently, the SEP mentions that servers must "validate this state somehow" to prevent modification. However, it doesn't explicitly address replay/hijacking attacks by different authenticated users even in the protocol requirements.

If request_state is offloaded to the client, what prevents User B (an authenticated attacker) from submitting a tools/call request using a stolen request_state from User A? If the blob only contains execution state and is not cryptographically bound to the specific Principal/User identity that initiated it, User B could effectively approve User A's requests.

Should the SEP explicitly mandate that request_state MUST include the identity of the initiator, and that servers MUST verify that the transport-level authenticated user matches the identity embedded in the request_state?

proposals/XXXX-MRTR.md
```diff
@@ -0,0 +1,1041 @@
...
+export interface JSONRPCRequest extends Request {
+  jsonrpc: typeof JSONRPC_VERSION;
+  id: RequestId;
+  // New field to carry the responses for the server's requests from the
+  // JSONRPCIncompleteResultResponse message.  For each key in the
+  // response's inputRequests field, the same key must appear here
+  // with the associated response.
+  inputResponses?: InputResponses;
+  // Request state passed back to the server from the client.
+  requestState?: string;
```
maxisbey, line 317: This isn't valid JSONRPC, these should be under a `params` field. See section 4 "Request": https://www.jsonrpc.org/specification

markdroth: Thanks for pointing that out!  I didn't realize that everything had to be under `params` here, nor that the responses have to be in the `result` field.  I've updated the proposal accordingly.

proposals/XXXX-MRTR.md
markdroth: Good point!  I've added a requirement about this in the "protocol requirements" section above.

@gjz22, please review this change and let me know what you think.  Thanks!

gjz22: LGTM maybe just replace "validate this state somehow" here with "validate this state as described in the protocol requirements"

markdroth: Done.

proposals/XXXX-MRTR.md
CaitieM20: @LucaButBoring could use some help clarifying the ask here. When you say consolidate on a single method.  I'm not sure what method that is.  Or are you suggesting:

Are you suggesting a separate  method to get the Input_Request Object and then a paired Input_Response method.  So something like:

`task/input_request` & `task/input_response`
I'm a bit confused by having a single method.  We have 2 separate calls here that have to happen to not require an SSE stream:
get the input_request currently on `task/result`
send the input_response currently a new method `task/input_response`

LucaButBoring: Explained on Discord but relaying here - I'm using `tasks/continue` as the suggested method name but we can name it whatever:

> 1. Client polls `tasks/get` and encounters `input_required` (request not rejected)
> 2. Client invokes `tasks/continue`; the request is rejected with the list of server-to-client requests that are necessary (as in ephemeral MRTR)
> 3. Client handles the requests by calling `tasks/continue` a second time with the responses; if this request succeeds then `tasks/continue` returns the new task state (same response shape as `tasks/get`)
> 4. Client eventually sees the `completed` status and calls `tasks/result` to retrieve the final value.

That makes the flow more consistent with ephemeral MRTR and gets rid of the premature `tasks/result` call, which has always been awkward.

proposals/XXXX-MRTR.md
```diff
@@ -0,0 +1,1179 @@
...
+   - Server may generate a large amount of state before requesting more
+     info from the client, and it may need to pick up that state to
+     continue processing after it receives the info from the client.
+   - Server may need to continue processing in the background while
+     waiting for more info from the client, in which case server-side
+     state is needed to track that ongoing processing.
+   - Examples: accessing an agent, spinning up a VM and needing user
+     interaction to manipulate the VM
+
+The vast majority of MCP tools will be ephemeral, and it is extremely
```
pcarleton, line 48: An alternative framing of this is that I think is consistent with the shift to the protocol being stateless is that the "persistent" cases are better suited to tasks w/ `input_required` than elicitation during a tool call.

edit: i see this called out below, might just be an earlier mention

proposals/XXXX-MRTR.md
```diff
@@ -0,0 +1,1179 @@
...
+For example, a server might send the following input requests:
+
+```json5
+"inputRequests": {
+  // Elicitation request.
+  "github_login": {
+    "method": "elicitation/create",
+    "params": {
+      "mode": "form",
+      "message": "Please provide your GitHub username",
```
pcarleton, line 188: nit: I think a stronger example here would help clarify what cases this flow is targeting.

e.g. an argument to the tool that's `github_username` would be a better fit.

edit: Okay I need to read further! The ADO one is a solid example, but does take more space to explain why it's good.  maybe we can come up with a more compact one that doesn't degenerate to tool arguments

proposals/XXXX-MRTR.md
```diff
@@ -0,0 +1,1179 @@
...
+
+#### Use Cases for Request State
+
+The "requestState" mechanism provides a mechanism for doing multiple
+round trips on the same logical request.  There are two main use-cases
+for this.
+
+##### Use Case 1: Rolling Upgrades
+
+Let's say that you are doing a rolling upgrade of your horizontally
```
pcarleton, line 621: would this case be useful for changing tool args as well?  not just changing your elicitations, but you could use an elicitation when an optional or new arg was left out (e.g. client had an outdated tool description).

proposals/XXXX-MRTR.md
pcarleton: The migration path for that might be "use tasks", which we could make easier.

felixweinberger: Chiming in here from the python & typescript sdk side of things; there are a couple of different things we might mean by "breaking things" here, and I think it's worth separating them:

### 1. If server or client implementers upgrade the SDK, will their code still compile/work?

I'd argue this is a non-goal. The code changes for tool authors are small and mechanical (see the migration demos in [typescript-sdk#1597](https://github.com/modelcontextprotocol/typescript-sdk/pull/1597) or even the python snippet right here). This is the kind of migration that can be documented in a `migration.md` and automated with AI tooling. Trying to keep this non-breaking adds permanent complexity to the SDKs for a one-time migration cost.

### 2. New client → old server: what happens?

This seems _mostly_ manageable. The client SDK keeps support for the existing SSE-based elicitation flow alongside the new MRTR path, and uses capability negotiation to determine which to use. It's a bit unfortunate to carry the legacy code, but it's contained to the client SDK (not tool authors) and can be deprecated on a timeline tied to a protocol version bump.

### 3. Old client → new server: what happens?

This is the hard one, and I think it's fundamentally unresolvable. The whole motivation for MRTR is enabling horizontally scaled servers without sticky sessions or long-lived SSE connections. An old client requires those things for elicitation - it expects the server to hold an SSE stream open and send elicitation requests on it. A new server that's deployed for horizontal scalability can't provide that by definition.

We could build a server-side SDK shim that translates new-style tool code back to the old SSE flow, _but that only works if the server can actually hold SSE connections_ - which kind of defeats the purpose of adopting MRTR in the first place. It makes maintenance of such servers _more_ complicated, not less.

I'd propose explicitly making graceful degradation the policy in this case: old clients can still connect to new servers and use tools normally, but elicitation and sampling aren't available until the client upgrades. The impact of this is scoped to tools that actually use server-initiated requests - I believe the vast majority of tools are simple request/response and would be completely unaffected.

cc: @pcarleton @maxisbey @pja-ant

proposals/XXXX-MRTR.md
```diff
@@ -0,0 +1,1179 @@
...
+If a tool implementation needs the client to respond to a set of
+input requests before it can even start processing but then later
+needs to do persistent processing, it can start using the ephemeral
+workflow and then switch to the persistent workflow by creating a task
+at that point.  This avoids the need for the server to store state until
+it actually has the information needed to start processing the request.
+This workflow would look like this:
+
+1. Client sends tool call request with task metadata.
+2. Server sends back `inputRequests` response indicating that more information is needed to process the request. This terminates the original request.
```
maxisbey, line 987: I'm wondering whether this is worth the potential added complexity for client side logic?

Without this change task augmented tool calls can already do elicitation/sampling in the new stateless transport. So, adding the ability for a task augmented tool call to return an `inputRequests` response doesn't gain any new functionality, it's only a potential optimization to avoid having to create a task before the initial required elicitation/sampling requests.

proposals/XXXX-MRTR.md
```diff
@@ -0,0 +1,1179 @@
...
+
+The persistent tool workflow will leverage Tasks. [`Tasks`](https://modelcontextprotocol.io/specification/draft/basic/utilities/tasks) already provide a mechanism to indicate that more information is needed to complete the request. The `input_required` Task Status allows the server to indicate that additional information is needed to complete processing the task.
+
+The workflow for `Tasks` is as follows:
+
+1. Server sets Task Status to `input_required`. The server can pause
+   processing the request at this point.
+2. Client retrieves the Task Status by calling `tasks/get` and sees that more information is needed.
+3. Client calls `task/result`
+4. Server returns the `InputRequests` object.
```
maxisbey, line 744: This is a breaking change to the tasks endpoint. [Currently](https://modelcontextprotocol.io/specification/draft/basic/utilities/tasks#result-retrieval):

> When a receiver receives a `tasks/result` request for a task in any other non-terminal status (`working` or `input_required`), it **MUST** block the response until the task reaches a terminal status.

Should probably be called out explicitly as a breaking change.

I'm also worried at some of these breaking changes around tasks being quite difficult to implement in the SDKs to support new/old server/client combinations. Especially making it so both the client and server is spec compliant in their interactions (i.e. we could just expose raw request methods and say "good luck" to sdk users but that's not fun haha).

Although, I suppose since we considered tasks to be experimental, we could just purposefully break it and say there's no backwards compatibility path? Either way I think it's worth noting that it's explicitly breaking.

proposals/XXXX-MRTR.md
```diff
@@ -0,0 +1,1179 @@
...
+
+The schema would look something like this:
+
+```typescript
+// Incomplete response.  May be sent in response to any client-initiated
+// request.
+// At least one of the the inputRequests and requestState fields must be
+// present, since the presence of these fields allow the client to
+// determine that the result is incomplete.
+export interface IncompleteResult extends Result {
```
maxisbey, line 298: Have you considered this instead being an JSONRPC error?

I am thinking this could potentially cause issues becaues this would mean that `tools/call` now returns multiple types for the same request.

The only case of this happening before is with task augmented tool calls, but importantly it was dependant on the type of request being sent out. A tool call would only send back a `CreateTaskResult` but _only_ if the request was task augmented as well. So in the SDKs this is easily separated out, and the result type to try and parse the response into is known ahead of time.

This will probably be a big issues for both the Typescript and Python SDKs which explicitly set the expected return type when making a request:
- Python does this on all requests sent out: https://github.com/modelcontextprotocol/python-sdk/blob/cb07adeca345effa111598889f6a8924a8722e6d/src/mcp/client/session.py#L313
- Same with Typescript: https://github.com/modelcontextprotocol/typescript-sdk/blob/3391bfb57fb05f07482480b83ec63b219d913292/packages/client/src/client/client.ts#L874

So by having this same request return two different types would beak this.

proposals/XXXX-MRTR.md
felixweinberger: +1 for using the experimental tag to explain a breaking change here.

proposals/XXXX-MRTR.md
markdroth: We'll definitely have to talk about how to handle this.

I don't think we really want this to be an error, because then we have to start extending the error type with individual fields.  Also, we want to use this same type in the task-based workflow for persistent tools, and an error doesn't really make sense in that context.

It's worth noting that for tasks, [SEP-2229](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2229) is proposing that the server be able to return a task to any call, even if the client does not explicitly ask for one.  So I think that even for that, the approach of determining the result schema based on the request is going to stop working.

proposals/XXXX-MRTR.md
markdroth: Yeah, I think that would work if the arg is optional or if there's no input schema.  If the arg is required and there is an input schema, then I think the SDK would reject the tool call request before the tool call impl has the opportunity to do this.  But presumably if you want to do this, you can just make the arg optional.

proposals/XXXX-MRTR.md
markdroth: @LucaButBoring may want to weigh in on this.

proposals/XXXX-MRTR.md
markdroth: I don't actually think this is going to add any significant complexity to client-side logic.  I think the ability to do this falls naturally out of the fact that we need to support both the ephemeral and persistent workflows.  Each retry attempt in the ephemeral workflow is essentially a completely new request, and the server can create a task in response to any such request.

Keep in mind that just because the client request indicates that it will accept a task doesn't mean that the server is required to start a task.  And also, [SEP-2229](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2229) is proposing simplifying this such that the client doesn't need to explicitly ask for a task in the first place; the server can just decide to send one whenever it needs to.

proposals/XXXX-MRTR.md
LucaButBoring: Yes, we'll need to make a breaking change to Tasks to support MRTR, and should call that out explicitly. I think the exact shape of the breaking change we make is still open to discussion, however - I plan to open a separate SEP for the relevant changes, which will probably involve (1) introducing a consolidated `tasks/continue` method that gets called repeatedly to advance the flow, and (2) making a breaking change to `tasks/result` to make it return an error if the task isn't complete, yet.

For now, can we simply state that MRTR will require breaking changes to Tasks, and that those will be addressed in a subsequent proposal? Let's describe the ephemeral/persistent split, delegate persistent use cases to Tasks, and explain the rationale (mostly as we're doing right now in the SEP), but remove the low-level details and example flow. I'll draft a TWG proposal by Wednesday addressing the full scope of the required changes.

markdroth: I don't think it makes sense to remove the tasks workflow from this SEP, because it's really one of the key workflows that we need to work here, and there are a lot of common data structures that we want for both.

Let's try to push forward the tasks simplification SEP in parallel with this one.  I'm perfectly happy to update this SEP to use whatever tasks workflow we converge on before it gets finalized.

proposals/XXXX-MRTR.md
ochafik: [SEP 1036](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1036) (implemented [here](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/887)) already introduced URLElicitationRequiredError, you'd maybe just need to update it to support form mode + tweak tool call semantics to allow resuming from an elicitation id?

markdroth: This SEP is explicitly removing `URLElicitationRequiredError` error.

markdroth:
Thas has now been published as [SEP-2322](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2322), so please move all further discussion there.  Thanks!

kurtisvg:
Merging this to keep a record of the original draft, similar to [IEFT Internet Drafts](https://www.ietf.org/participate/ids/). Future updates and comments should be directed to [SEP-2322](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2322).

kurtisvg (approved):

proposals/XXXX-MRTR.md
LucaButBoring: Published: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2339

PR #15, opened 2026-02-10
SEP-XXXX HTTP Standardization

mikekistler:
## Motivation and Context

This PR introduces a SEP for incorporating HTTP features into the HTTP Transport that support processing by "middle boxes" on the internet, for tasks like routing, tracing, and priorization.

## How Has This Been Tested?

I have a reference implementation built on the C# SDK that illustrates the basic features in this SEP.

https://github.com/mikekistler/csharp-sdk/tree/mdk/http-standardization

## Breaking Changes

All changes are gated by protocol version and thus not breaking.

## Types of changes

- [ ] Documentation update

## Checklist

- [ ] New and existing tests pass locally
- [ ] I have added appropriate error handling
- [ ] I have added or updated documentation as needed

## Additional context

This PR is primarily to allow comment and discussion within the Transports Work Group.

### DIFF
--- a/proposals/xxxx-http-standardization.md
+++ b/proposals/xxxx-http-standardization.md
@@ -241,7 +241,32 @@ The prefix `=?base64?` and suffix `?=` indicate that the value is Base64-encoded
 
 **Server Validation**
 
-Servers MUST validate that encoded header values, after decoding if Base64-encoded, match the corresponding values in the request body. Servers MUST reject requests with a `400 Bad Request` status if:
+Servers MUST validate that encoded header values, after decoding if Base64-encoded, match the corresponding values in the request body. Servers MUST reject requests with a `400 Bad Request` HTTP status if any validation fails.
+
+**Error Code**
+
+When rejecting a request due to header validation failure, servers MUST return a JSON-RPC error response with the following error code:
+
+| Code | Name | Description |
+|------|------|-------------|
+| `-32001` | `HeaderMismatch` | The HTTP headers do not match the corresponding values in the request body, or required headers are missing/malformed. |
+
+This error code is in the JSON-RPC implementation-defined server error range (`-32000` to `-32099`).
+
+**Error Response Format**:
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "error": {
+    "code": -32001,
+    "message": "Header mismatch: Mcp-Tool-Name header value 'foo' does not match body value 'bar'"
+  }
+}
+```
+
+**Validation Failure Conditions**:
 
 - A required header is missing
 - A header value does not match the request body value
@@ -521,9 +546,9 @@ This section defines edge cases that conformance tests MUST cover to ensure inte
 
 | Test Case | Header Value | Body Value | Expected Behavior |
 |-----------|--------------|------------|-------------------|
-| Method mismatch | `Mcp-Method: tools/call` | `"method": "prompts/get"` | Server MUST reject with 400 |
-| Tool name mismatch | `Mcp-Tool-Name: foo` | `"params": {"name": "bar"}` | Server MUST reject with 400 |
-| Missing required header | (no `Mcp-Method`) | Valid body | Server MUST reject with 400 |
+| Method mismatch | `Mcp-Method: tools/call` | `"method": "prompts/get"` | Server MUST reject with 400 and error code `-32001` |
+| Tool name mismatch | `Mcp-Tool-Name: foo` | `"params": {"name": "bar"}` | Server MUST reject with 400 and error code `-32001` |
+| Missing required header | (no `Mcp-Method`) | Valid body | Server MUST reject with 400 and error code `-32001` |
 | Extra whitespace in header | `Mcp-Tool-Name:  foo ` | `"params": {"name": "foo"}` | Server MUST accept (trim whitespace per HTTP spec) |
 
 #### Special Characters in Values
@@ -591,8 +616,8 @@ This section defines edge cases that conformance tests MUST cover to ensure inte
 | Test Case | Header Value | Expected Behavior |
 |-----------|--------------|-------------------|
 | Valid Base64 | `=?base64?SGVsbG8=?=` | Server decodes to `"Hello"` and validates |
-| Invalid Base64 padding | `=?base64?SGVsbG8?=` | Server MUST reject with 400 |
-| Invalid Base64 characters | `=?base64?SGVs!!!bG8=?=` | Server MUST reject with 400 |
+| Invalid Base64 padding | `=?base64?SGVsbG8?=` | Server MUST reject with 400 and error code `-32001` |
+| Invalid Base64 characters | `=?base64?SGVs!!!bG8=?=` | Server MUST reject with 400 and error code `-32001` |
 | Missing prefix | `SGVsbG8=` | Server treats as literal value, not Base64 |
 | Missing suffix | `=?base64?SGVsbG8=` | Server treats as literal value, not Base64 |
 | Malformed wrapper | `=?BASE64?SGVsbG8=?=` | Server MUST accept (case-insensitive prefix) |
--- a/proposals/xxxx-http-standardization.md
+++ b/proposals/xxxx-http-standardization.md
@@ -174,8 +174,66 @@ When constructing a `tools/call` request via HTTP transport, the client:
 
 1. Inspects the tool's `inputSchema` for properties marked with `x-mcp-header`
 2. Extracts the value provided for that parameter
-3. Sanitizes the value to ensure it is safe for HTTP headers (ASCII only, no newlines)
+3. Encodes the value according to the rules in [Value Encoding](#value-encoding)
 4. Appends a header to the request: `Mcp-Param-{Name}: {Value}`
+
+#### Value Encoding
+
+Clients MUST encode parameter values before including them in HTTP headers to ensure safe transmission and prevent injection attacks.
+
+**Character Restrictions**
+
+Per [RFC 9110](https://datatracker.ietf.org/doc/html/rfc9110#name-field-values), HTTP header field values must consist of visible ASCII characters (0x21-0x7E), space (0x20), and horizontal tab (0x09). The following characters are explicitly prohibited:
+
+- Carriage return (`\r`, 0x0D)
+- Line feed (`\n`, 0x0A)
+- Null character (`\0`, 0x00)
+- Any character outside the ASCII range (> 0x7F)
+
+**Encoding Rules**
+
+Clients MUST apply the following encoding rules in order:
+
+1. **Type conversion**: Convert the parameter value to its string representation:
+   - `string`: Use the value as-is
+   - `number`: Convert to decimal string representation (e.g., `42`, `3.14`)
+   - `boolean`: Convert to lowercase `"true"` or `"false"`
+
+2. **ASCII validation**: Check if the string contains only valid ASCII characters (0x20-0x7E):
+   - If valid, proceed to step 3
+   - If invalid (contains non-ASCII or control characters), apply Base64 encoding (see below)
+
+3. **Control character check**: If the string contains any control characters (0x00-0x1F or 0x7F):
+   - Apply Base64 encoding (see below)
+
+4. **Length validation**: If the encoded value exceeds 8192 bytes, the client MUST omit the header and MAY log a warning
+
+**Base64 Encoding for Unsafe Values**
+
+When a value cannot be safely represented as a plain ASCII header value, clients MUST use Base64 encoding with the following format:
+
+```
+Mcp-Param-{Name}: =?base64?{Base64EncodedValue}?=
+```
+
+The prefix `=?base64?` and suffix `?=` indicate that the value is Base64-encoded. Servers and intermediaries that need to inspect these values MUST decode them accordingly.
+
+**Example**:
+
+A parameter value of `"Hello, 世界"` (contains non-ASCII) would be encoded as:
+
+```
+Mcp-Param-Greeting: =?base64?SGVsbG8sIOS4lueVjA==?=
+```
+
+**Server Validation**
+
+Servers MUST validate that encoded header values, after decoding if Base64-encoded, match the corresponding values in the request body. Servers MUST reject requests with a `400 Bad Request` status if:
+
+- A required header is missing
+- A header value does not match the request body value
+- A Base64-encoded value cannot be decoded
+- A header value contains invalid characters
 
 #### Example: Geo-Distributed Database
 
@@ -407,18 +465,30 @@ This is a new, optional feature. Existing tools without `x-mcp-header` propertie
 
 ### Header Injection
 
-Clients MUST sanitize parameter values before including them in headers to prevent header injection attacks:
+Header injection attacks occur when malicious values containing control characters (especially `\r\n`) are included in headers, potentially allowing attackers to inject additional headers or terminate the header section early.
 
-- Remove or reject values containing newline characters (`\r`, `\n`)
-- Ensure values contain only valid ASCII characters for HTTP header values
+Clients MUST follow the [Value Encoding](#value-encoding) rules defined in this specification. These rules ensure that:
+
+- Control characters are never included in header values
+- Non-ASCII values are safely encoded using Base64
+- Values exceeding safe length limits are omitted
 
 ### Header Spoofing
 
 Servers MUST validate that header values match the corresponding values in the request body. This prevents clients from sending mismatched headers to manipulate routing while executing different operations.
 
+For example, a malicious client could attempt to:
+- Route a request to a less-secured region while executing operations intended for a high-security region
+- Bypass rate limits by spoofing tenant identifiers
+- Evade security policies by misrepresenting the operation being performed
+
 ### Information Disclosure
 
-Tool parameter values designated for headers will be visible to network intermediaries. Server developers SHOULD NOT mark sensitive parameters (passwords, tokens, PII) with `x-mcp-header`.
+Tool parameter values designated for headers will be visible to network intermediaries (load balancers, proxies, logging systems). Server developers:
+
+- SHOULD NOT mark sensitive parameters (passwords, API keys, tokens, PII) with `x-mcp-header`
+- SHOULD document which parameters are exposed as headers
+- SHOULD consider that Base64 encoding provides no confidentiality—it is merely an encoding, not encryption
 
 ## Reference Implementation
 
@@ -427,5 +497,5 @@ _To be provided before this SEP reaches Final status._
 Implementation requirements:
 
 - **Server SDKs**: Provide a mechanism (attribute/decorator) for marking parameters with `x-mcp-header`
-- **Client SDKs**: Implement the client behavior for extracting and sanitizing header values
+- **Client SDKs**: Implement the client behavior for extracting and encoding header values
 - **Validation**: Both sides must validate header/body consistency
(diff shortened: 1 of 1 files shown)


proposals/xxxx-http-standardization.md
```diff
@@ -0,0 +1,755 @@
...
+
+## Abstract
+
+This SEP proposes exposing critical routing and context information in standard HTTP header locations for the Streamable HTTP transport. By mirroring key fields from the JSON-RPC payload into HTTP headers, network intermediaries such as load balancers, proxies, and observability tools can route and process MCP traffic without deep packet inspection, reducing latency and computational overhead.
+
+## Motivation
+
+Current MCP implementations over HTTP bury all routing information within the JSON-RPC payload. This creates friction for network infrastructure:
+
+- **Load balancers** must terminate TLS and parse the entire JSON body to extract routing information (e.g., region, tool name)
```
CaitieM20, line 21: Removing Region since that's a custom param. I think it'll be easier for people to digest by starting with listing the Protocol specifics. 
```suggestion
- **Load balancers** must terminate TLS and parse the entire JSON body to extract routing information (e.g., method, tool name, etc...)
```

proposals/xxxx-http-standardization.md
```diff
@@ -0,0 +1,755 @@
...
+    "capabilities": {},
+    "clientInfo": {
+      "name": "ExampleClient",
+      "version": "1.0.0"
+    }
+  }
+}
+```
+
+### Custom Headers from Tool Parameters
```
CaitieM20, line 136: I'm very supportive of the standard headers.

I think we should vote on whether to split custom headers out, in past WG discussions we noted that this adds additional client implementation complexity.

proposals/xxxx-http-standardization.md
```diff
@@ -0,0 +1,755 @@
...
+{
+  "name": "execute_sql",
+  "description": "Execute SQL on Google Cloud Spanner",
+  "inputSchema": {
+    "type": "object",
+    "properties": {
+      "region": {
+        "type": "string",
+        "description": "The region to execute the query in",
+        "x-mcp-header": "Region"
```
CaitieM20, line 167: not a super strong opinion but having transport concerns embedded within the tool definition object feels like a mixing of concerns (tool definition, optional transport considerations).

proposals/xxxx-http-standardization.md
```diff
@@ -0,0 +1,755 @@
...
+Custom headers (those defined via `x-mcp-header`) follow different rules than standard headers:
+
+| Scenario | Client Behavior | Server Behavior |
+|----------|-----------------|-----------------|
+| Parameter value provided | Client MUST include the header | Server MUST validate header matches body |
+| Parameter value is `null` | Client MUST omit the header | Server MUST NOT expect the header |
+| Parameter not in arguments | Client MUST omit the header | Server MUST NOT expect the header |
+| Client omits header but value is in body | Non-conforming client | Server SHOULD accept the request but MAY log a warning |
+
+**Rationale for Lenient Server Behavior**: Unlike standard headers, custom headers are primarily intended to enable infrastructure optimizations (routing, rate limiting, etc.). A missing custom header does not prevent the server from processing the request—the parameter value is still available in the request body. Servers SHOULD prioritize interoperability with older or non-conforming clients over strict enforcement. However, servers MAY choose to reject such requests in environments where header-based routing is critical to correct operation.
```
CaitieM20: I don't think I follow, aren't standard & custom headers just about infra optimizations.

proposals/xxxx-http-standardization.md
```diff
@@ -0,0 +1,755 @@
...
+```
+
+Despite this additional complexity in some frameworks, header-based routing was chosen because:
+
+1. **Infrastructure benefits outweigh framework complexity**: The primary goal is enabling network infrastructure (load balancers, proxies, WAFs) to route and process requests without body parsing. This benefit applies regardless of the server framework.
+
+2. **Single endpoint simplicity**: Keeping a single `/mcp` endpoint simplifies CORS configuration, authentication middleware, and client implementation.
+
+3. **Separation of concerns**: The URL path identifies the MCP service; headers identify the operation within that service. This mirrors how HTTP methods (GET, POST, etc.) work alongside paths.
+
```
CaitieM20, line 520:
```suggestion
4. **Backwards Compatibility** introducing path based routing would require all existing MCP Servers to take a major update, and potentially support two sets of endpoints to support multiple versions. Even if the SDKs can paper over this additional operational concerns like testing, metrics, etc would need to happen. Header based routing requires minimal client side changes. And clients which don't opt in will still function correctly.
```

proposals/xxxx-http-standardization.md
```diff
@@ -0,0 +1,755 @@
...
+2. **Established pattern**: JSON Schema explicitly supports extension keywords (properties starting with `x-`), and this pattern is widely used in ecosystems like OpenAPI. Tool authors and SDK developers are already familiar with this approach.
+
+3. **Schema composability**: When schemas are composed, extended, or referenced using `$ref`, the `x-mcp-header` annotation travels with the property definition. A separate metadata structure would require complex synchronization logic to maintain consistency.
+
+4. **Tooling compatibility**: Existing JSON Schema validators ignore unknown keywords by default, so adding `x-mcp-header` doesn't break existing schema validation. Tools that don't understand this extension simply skip it.
+
+5. **Reduced complexity**: A separate metadata structure would require defining a mapping mechanism (e.g., JSON Pointer or property paths) to associate headers with properties, adding implementation complexity and potential for errors.
```
CaitieM20, line 551: I'm going to push us to get more explicit on what Complexity we are reducing. By having the Tool Definition and Transport Specific data merged in the same object. This kind of breaks the abstraction boundary for me. 

Seems like this is saying its reducing client implementation complexity. I'd be curious if our SDK maintainers/client implementors could weigh in here.  Is this the right tradeoff here.

proposals/xxxx-http-standardization.md
```diff
@@ -0,0 +1,755 @@
...
+| String with newline | `"line1\nline2"` | `Mcp-Param-Text: =?base64?bGluZTEKbGluZTI=?=` |
+| String with carriage return | `"line1\r\nline2"` | `Mcp-Param-Text: =?base64?bGluZTENCmxpbmUy?=` |
+| String with leading tab | `"\tindented"` | `Mcp-Param-Text: =?base64?CWluZGVudGVk?=` |
+| Empty string | `""` | `Mcp-Param-Name: ` (empty value) |
+| String exceeding length limit | 10000 character string | Client MUST omit header |
+
+#### Type Restriction Violations
+
+| Test Case | Property Type | x-mcp-header Present | Expected Behavior |
+|-----------|---------------|---------------------|-------------------|
```
CaitieM20, line 719: So if a Client receives a tool definition with a x-mcp-header defined for one of these types what should its behavior be? This assumes an MCP Server has been misconfigured. 

Does it just omit the x-mcp-header?
Does it refuse to call the tool?

What's the recommended behavior.

proposals/xxxx-http-standardization.md
```diff
@@ -0,0 +1,755 @@
...
+
+#### Schema Extension
+
+The `x-mcp-header` property specifies the name portion used to construct the header name `Mcp-Param-{name}`.
+
+**Constraints on `x-mcp-header` values**:
+
+- MUST contain only ASCII characters (excluding space and `:`)
+- MUST be case-insensitively unique among all `x-mcp-header` values in the `inputSchema`
+- MUST only be applied to parameters with primitive types (number, string, boolean)
```
CaitieM20, line 151: I get why the x-mcp-header should only be valid on these fields. For our SDK maintainers does this introduce any challenges or complexity in adding this in to some param types and not others?

CaitieM20 (commented):
Supportive of the Standard Headers. I think there is a discussion/vote to have on whether we want to bring in Custom Headers now as well given the number of edge cases.

proposals/xxxx-http-standardization.md
kurtisvg, line 151: If we are base64 encoding, why do we need to limit to primitive types?

proposals/xxxx-http-standardization.md
```diff
@@ -0,0 +1,755 @@
...
+3. Encodes the value according to the rules in [Value Encoding](#value-encoding)
+4. Appends a header to the request: `Mcp-Param-{Name}: {Value}`
+
+#### Value Encoding
+
+Clients MUST encode parameter values before including them in HTTP headers to ensure safe transmission and prevent injection attacks.
+
+**Character Restrictions**
+
+Per [RFC 9110](https://datatracker.ietf.org/doc/html/rfc9110#name-field-values), HTTP header field values must consist of visible ASCII characters (0x21-0x7E), space (0x20), and horizontal tab (0x09). The following characters are explicitly prohibited:
```
kurtisvg: What should a client do if they are specified?

proposals/xxxx-http-standardization.md
```diff
@@ -0,0 +1,755 @@
...
+   - Apply Base64 encoding (see below)
+
+3. **ASCII validation**: Check if the string contains only valid ASCII characters (0x20-0x7E):
+   - If valid, proceed to step 4
+   - If invalid (contains non-ASCII characters), apply Base64 encoding (see below)
+
+4. **Control character check**: If the string contains any control characters (0x00-0x1F or 0x7F):
+   - Apply Base64 encoding (see below)
+
+5. **Length validation**: If the encoded value exceeds 8192 bytes, the client MUST omit the header and MAY log a warning
```
kurtisvg: Not sure I love this. I understand the rationale, but I think it's presumptive:

Here's some common limits:
- Apache: The default is typically 8 KB (8190 bytes) for each header field.
- Nginx: Defaults range from 4 KB to 8 KB, based on system page size.
- IIS (Microsoft): Limits vary by version but are often in the 8 KB to 16 KB range for the request line and headers combined.
- Tomcat: Defaults are typically 8 KB to 16 KB.
- Cloudflare (CDN): Supports a total limit of up to 128 KB for both request and response headers.

With the current value we're still causing issues with the smaller header limits, but also significantly restricting the larger ones. 

IMO, this **should** be service dependent. Since the service specifies the headers they should know how long it should be.

proposals/xxxx-http-standardization.md
kurtisvg: My 2c is these headers should be treated like the standard ones. The server should be allowed to reject the request if the header is missing, otherwise we're forcing different steps to implement validation logic.

proposals/xxxx-http-standardization.md
```diff
@@ -0,0 +1,755 @@
...
+| Original Value | Reason | Encoded Header Value |
+|----------------|--------|---------------------|
+| `"us-west1"` | Plain ASCII | `Mcp-Param-Region: us-west1` |
+| `"Hello, 世界"` | Contains non-ASCII | `Mcp-Param-Greeting: =?base64?SGVsbG8sIOS4lueVjA==?=` |
+| `" padded "` | Leading/trailing spaces | `Mcp-Param-Text: =?base64?IHBhZGRlZCA=?=` |
+| `"line1\nline2"` | Contains newline | `Mcp-Param-Text: =?base64?bGluZTEKbGluZTI=?=` |
+
+**Server Validation**
+
+Servers MUST validate that encoded header values, after decoding if Base64-encoded, match the corresponding values in the request body. Servers MUST reject requests with a `400 Bad Request` HTTP status if any validation fails.
```
kurtisvg: Should we clarify this that servers means "any step in the routing chain where the body is inspected"

proposals/xxxx-http-standardization.md
kurtisvg: I disagree on the abstraction boundary -- it's an attribute of the parameter (much like type) and (imo) makes sense to put it near the parameter.  

I also think that it's easier on the server implementation to put this in the same location.

proposals/xxxx-http-standardization.md
```diff
@@ -0,0 +1,755 @@
...
+| Resource URI with special chars | `file:///path/to/file%20name.txt` | Client sends as-is; server accepts |
+| Resource URI with query string | `https://example.com/resource?id=123` | Client sends as-is; server accepts |
+
+### Custom Header Edge Cases
+
+#### x-mcp-header Name Conflicts
+
+| Test Case | Schema | Expected Behavior |
+|-----------|--------|-------------------|
+| Duplicate header names (same case) | Two properties with `"x-mcp-header": "Region"` | Server MUST reject tool definition |
```
kurtisvg: How does a SERVER reject the tool definition? isnt the server that serves this?

proposals/xxxx-http-standardization.md
```diff
@@ -0,0 +1,755 @@
...
+| Duplicate header names (different case) | `"x-mcp-header": "Region"` and `"x-mcp-header": "REGION"` | Server MUST reject tool definition (case-insensitive uniqueness) |
```
kurtisvg: nit: headers are supposed to be case in-sensitive. Should we just sanitize all values to lowercase?

proposals/xxxx-http-standardization.md
kurtisvg: +1

proposals/xxxx-http-standardization.md
```diff
@@ -0,0 +1,755 @@
...
+    elif method == 'resources/read':
+        return handle_resources_read(request)
+    # ... etc
+```
+
+Despite this additional complexity in some frameworks, header-based routing was chosen because:
+
+1. **Infrastructure benefits outweigh framework complexity**: The primary goal is enabling network infrastructure (load balancers, proxies, WAFs) to route and process requests without body parsing. This benefit applies regardless of the server framework.
+
+2. **Single endpoint simplicity**: Keeping a single `/mcp` endpoint simplifies CORS configuration, authentication middleware, and client implementation.
```
kurtisvg: Not sure I buy this. These things all work fine with multiple endpoints. (in fact some thing like rate limiting would work better with path based stuff)

proposals/xxxx-http-standardization.md
kurtisvg: Presumably this is probably number 1 reason

proposals/xxxx-http-standardization.md
```diff
@@ -0,0 +1,755 @@
...
+
+3. **Separation of concerns**: The URL path identifies the MCP service; headers identify the operation within that service. This mirrors how HTTP methods (GET, POST, etc.) work alongside paths.
```
kurtisvg: I don't really buy this one either. Normal HTTP services use different paths.

proposals/xxxx-http-standardization.md
mikekistler: Leaving this as is since the workgroup has agreed to keep this as one SEP at least for now.

proposals/xxxx-http-standardization.md
mikekistler: We'll keep it as one SEP for now and see how this is received by the core maintainers.

proposals/xxxx-http-standardization.md
mikekistler: Understood. I think it is simpler this way, but maybe that's because of my years of experience with OpenAPI and how extension properties are used for all kinds of things in that world (check out [this compilation of extensions done in 2021](https://github.com/Mermade/openapi-specification-extensions/blob/main/extensions/2021/extensions.tsv)).

Let's leave it this way for now, but if the consensus in the core maintainers meeting is to spilit this out we can do that.

proposals/xxxx-http-standardization.md
mikekistler: There is certainly some additional logic needed for this restriction, but in the C# SDK it is relatively straightforward. This code in the schema customization logic throws an exception when a non-primitive type is annotated to be mirrored into a header.

```csharp
            // Validate parameter type is primitive (string, number, boolean)
            var paramType = Nullable.GetUnderlyingType(param.ParameterType) ?? param.ParameterType;
            if (!IsPrimitiveType(paramType))
            {
                throw new ArgumentException(
                    $"Parameter '{param.Name}' with McpHeaderAttribute must be a primitive type (string, number, or boolean), " +
                    $"but was '{param.ParameterType.Name}'.");
            }
```

where

```csharp
    private static bool IsPrimitiveType(Type type) =>
        type == typeof(string) ||
        type == typeof(bool) ||
        type == typeof(byte) ||
        type == typeof(sbyte) ||
        type == typeof(short) ||
        type == typeof(ushort) ||
        type == typeof(int) ||
        type == typeof(uint) ||
        type == typeof(long) ||
        type == typeof(ulong) ||
        type == typeof(float) ||
        type == typeof(double) ||
        type == typeof(decimal);
```

I would be surprised if presented a challenge in any other languages.

mikekistler: Because base64 encoding is a mapping from binary -> string, and for non-primitive types we'd still need to define the mapping for "instance" -> binary. I acknowledge that this _could_ be done, but I'm waiting for the clear requirement that it is needed before adding that complexity.

proposals/xxxx-http-standardization.md
mikekistler: A "client" (the originator of the request) should use the serialization rules defined in this SEP so that it never sends a header with invalid characters. A receiver of these requests should  reject / fail any request with a header containing invalid characters. I think this is already required by the HTTP RFCs but we could state it here explicitly if you think that is necessary / desirable.

proposals/xxxx-http-standardization.md
mikekistler: Does this mean that you want the client to not impose any limitation on the size of the encoded value?

proposals/xxxx-http-standardization.md
mikekistler: I think this is from the perspective of the "server SDK". Perhaps we should reword to state that an MCP server should not expose a tool definition where two properties of the input schema have the same `x-mcp-header` value (case-insensitively). Would that be clearer?

proposals/xxxx-http-standardization.md
mikekistler: Sanitizing the values to lowercase before checking for unique values is _an implementation_ that ensures this edge case is covered.

proposals/xxxx-http-standardization.md
mikekistler: I'll update the SEP to address this and resolve this comment when the update is pushed.

proposals/xxxx-http-standardization.md
mikekistler: Looking at this again, I think this and the other edge cases in this section should say "Client" rather than "Server". I will also add a clear requirement on the client to perform this validation.

proposals/xxxx-http-standardization.md
mikekistler: Again, I am tainted by the widespread practice of using extension properties for all manner of purposes in OpenAPI. And because extension properties are already possible and supported, this approach requires no schema changes whatsoever.

But I don't feel strongly about this. I doubt this will impact server or client developers either way if they use an SDK, since I expect the SDKs will expose this as an attribute/decorator on the tool parameter regardless of how it's represented in the tool definition.

Let's see if there is any push-back from the core maintainers and if so we can change this to use a new metadata field instead.

proposals/xxxx-http-standardization.md
kurtisvg: Yes. I think the server should do this. There is already a dedicated HTTP code, so I don't really think it's going to be a problem to diagnose.

kurtisvg (approved):

markdroth (approved):
This looks great!

CaitieM20:
Closing since this has moved to an official SEP (https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2243)

PR #20, opened 2026-02-24
SEP-XXXX: Data-Layer Sessions for MCP

evalstate:
Very early draft, will demo and tidy-up for feedback/development.

## Types of changes

- [ ] New feature (non-breaking change which adds functionality)
- [ ] Documentation update

## Checklist

- [ ] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [ ] My code follows the repository's style guidelines
- [ ] New and existing tests pass locally
- [ ] I have added appropriate error handling
- [ ] I have added or updated documentation as needed


proposals/data-layer-sessions-api-shapes.jsonc
```diff
@@ -0,0 +1,210 @@
...
+// Request
+{
+  "jsonrpc": "2.0",
+  "id": 4,
+  "method": "tools/call",
+  "params": {
+    "name": "notebook_append",
+    "arguments": { "text": "remember this" },
+    "_meta": {
+      "mcp/session": { "id": "sess-a1b2c3d4e5f6" }
```
gjz22: If you want an expiry, maybe, the client needs to pass back the original expiry to mitigate clock skew issues?

ggoodman:
This shifts session creation from the server to the client. I can understand why clients might want some measure of control over this.

But what I feel might be missing is a server-sent hint or error-type that would _require_ the client to operate within a session. Previously, that's been a pretty janky [dance](https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#session-management):

> When a client receives HTTP 404 in response to a request containing an MCP-Session-Id, it MUST start a new session by sending a new InitializeRequest without a session ID attached.

Can we consider giving a way for servers to signal the requirement to use sessions?

proposals/data-layer-sessions-api-shapes.jsonc
evalstate: Expiry is intended as a moderately strong UX hint rather than a guarantee either way. I'd like to avoid handling clock skew here if we can!

evalstate:
> This shifts session creation from the server to the client. I can understand why clients might want some measure of control over this.
> 
> But what I feel might be missing is a server-sent hint or error-type that would _require_ the client to operate within a session. Previously, that's been a pretty janky [dance](https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#session-management):
> 
> > When a client receives HTTP 404 in response to a request containing an MCP-Session-Id, it MUST start a new session by sending a new InitializeRequest without a session ID attached.
> 
> Can we consider giving a way for servers to signal the requirement to use sessions?

Overall the stateless transport design is preferring optimistic usage (e.g. try and fail). I made a small update that the recommended thing to do would be to call `discover` and find out that way.

I do think that there may be situations where sessions are optional - which in this current design would be indicated by the server supporting the capability, and responding to `discover` affirmatively when no `SessionMetadata` is passed.  

Tool annotations are slightly tricker is the `tools/list` mechanism may itself be gated by a session requirement.

proposals/0000-data-layer-sessions.md
```diff
@@ -0,0 +1,427 @@
...
+ 
+This proposal adopts a similar pattern (server-issued opaque tokens that clients return unmodified) but implements it in the JSON-RPC message layer rather than HTTP Headers, enabling consistent session semantics across non-HTTP transports.
+
+### Session Update Sequencing
+
+There are no ordering guarantees for requests/responses, meaning a Last-Write-Wins strategy by default. Servers should be aware of this potential race condition and include appropriate mitigations if needed.
+
+A future design may introduce a monotonic state sequence to allow the client to identify the ordering of state content. 
+
+Servers that have this as a critical requirement should manage session state at the server.
```
kurtisvg: nit: suggest updating for clarity. I'm not sure we need a monotonic state sequence. 
```suggestion
Because MCP clients may execute multiple requests concurrently (e.g., parallel tool calling), there are no inherent ordering guarantees for request/response cycles. This creates a Last-Write-Wins race condition if the server relies entirely on the client-echoed state token to manage highly mutable data.

To resolve this, servers dealing with concurrent mutations SHOULD NOT rely on the state token. Instead, the server SHOULD use the sessionId as a lookup key for a server-side state management mechanism (e.g., a database, cache, or in-memory store).

Using server-side state naturally delegates concurrency control to the server environment, keeping the MCP client implementation simple and eliminating the need for sequence tracking within the protocol layer. The state token remains available strictly for simple, stateless server deployments where concurrent mutations are not expected.
```

proposals/0000-data-layer-sessions.md
```diff
@@ -0,0 +1,427 @@
...
+> **Status:** Early Draft  
+> **Date:** 2026-02-23  
+> **Track:** transport-wg/sessions  
+> **Author(s):** Shaun Smith
+
+## Abstract
+
+This proposal introduces application level sessions within the MCP Data Layer. Sessions are created by the Client, and allow the Server to store an opaque state token.
+
+This proposal should be reviewed alongside SEP-1442, and assumes that the `initialize` operation and associated StreamableHTTP session creation is deprecated.
```
kurtisvg:
```suggestion
This proposal should be reviewed alongside SEP-1442. It assumes that the legacy initialize operation—and its side effect of implicit transport-level session creation (e.g., in Streamable HTTP)—is deprecated in favor of a stateless capability discovery mechanism (e.g., a /discover endpoint).

Consequently, per SEP-1442, if a client attempts to invoke sessions/create or utilize session metadata on a server that does not support this extension, the server MUST reject the request with a standard JSON-RPC -32601 Method not found error.
```

proposals/0000-data-layer-sessions.md
```diff
@@ -0,0 +1,427 @@
...
+
+The Client **MUST** securely associate retained sessions with the issuing Server. The Client will typically establish identity through a mixture of _connection target_ and _user identity_. 
+
+`expiresAt` is a hint, and may be updated by the Server in future responses. The Host **MAY** use the `expiresAt` to indicate potentially stale sessions to the User. 
+
+`state` **MUST** be retained by the Client and sent with future requests for that session.
+
+#### Using Sessions
+
+To use a Session the Client request includes SessionMetadata in `_meta["io.modelcontextprotocol/session"]`:
```
kurtisvg, line 127: Do we need to call out how notifications work with sessions? 

Specifically the server/listen request a client sends should have sessions ids? Maybe more then one? that let is subscribe to notifications from the server for that session(s). The notifications should have the session metadata attached, or no session if they are broadly applicable

proposals/0000-data-layer-sessions.md
```diff
@@ -0,0 +1,427 @@
...
+  }
+}
+```
+
+1. The Client **MUST** update the `state` value if updated by the Server. See note on [Ordering](#session-update-sequencing))
+1. The Client **MAY** update the `expiresAt` value if updated by the Server.
+
+#### Deleting Sessions
+
+Clients **SHOULD** delete sessions that are no longer required to allow the Server to reclaim unneeded resources.
```
kurtisvg: Should we clarify that servers should clean up sessions after a while?

proposals/0000-data-layer-sessions.md
```diff
@@ -0,0 +1,427 @@
...
+
+#### Errors
+
+```json
+ {
+   "jsonrpc": "2.0",
+   "id": 42,
+   "error": {
+     "code": -32043,
+     "message": "Session not found",
```
kurtisvg, line 361: Should we add a `SESSION_REQUIRED` error? to indicate that a tool requires a session but doesn't have one

kurtisvg (changes requested):

proposals/0000-data-layer-sessions.md
markdroth, line 127: If we want sessions to be more of a first-class thing, should this be a new top-level field in params rather than being in metadata?

proposals/0000-data-layer-sessions.md
```diff
@@ -0,0 +1,427 @@
...
+`expiresAt` is a hint, and may be updated by the Server in future responses. The Host **MAY** use the `expiresAt` to indicate potentially stale sessions to the User. 
+
+`state` **MUST** be retained by the Client and sent with future requests for that session.
+
+#### Using Sessions
+
+To use a Session the Client request includes SessionMetadata in `_meta["io.modelcontextprotocol/session"]`:
+
+1. The Server MUST treat that sessionId as the session context for processing the request.
+1. Succesful responses MUST include \_meta["io.modelcontextprotocol/session"].
```
markdroth: Why do we need the server to echo back the session ID in every response?  The client has already created the session and included it in the request, so why do we need the server to echo it back?

I think it makes sense for the server to be able to do this when it's useful to do so (e.g., to change the state or the expiration time).  But if the server doesn't actually have any new information to communicate, it seems a little wasteful to include this information in every single response.

proposals/0000-data-layer-sessions.md
```diff
@@ -0,0 +1,427 @@
...
+     "code": -32043,
+     "message": "Session not found",
+     "data": {
+       "sessionId": "sess-a1b2c3d4e5f6"
+     }
+   }
+ }
+```
+
+1. Unknown sessions **MUST** result in a  `-32043 SESSION_NOT_FOUND` Error. The Client **SHOULD** treat the Session as permanently invalidated.
```
markdroth: Why is this a MUST?  What if the server has some calls that need session data and others that don't?  It seems like if a call really doesn't happen to interact with session data in any way, the server shouldn't be required to fail that call just because the session is bad.

proposals/0000-data-layer-sessions.md
```diff
@@ -0,0 +1,427 @@
...
+### Data Types
+
+#### SessionMetadata
+
+_The following notes on sessionId are taken from the existing Streamable HTTP Transport guidance._
+
+**sessionId:**
+1. The `sessionId` **SHOULD** be globally unique and cryptographically secure (e.g., a securely generated UUID, a JWT, or a cryptographic hash).
+1. The `sessionId` MUST only contain visible ASCII characters (ranging from 0x21 to 0x7E).
+1. The client MUST handle the `sessionId` in a secure manner, see Session Hijacking mitigations for more details. (
```
markdroth: Trailing `(` at the end of this line.  Was that intended to be a link to something?

proposals/0000-data-layer-sessions.md
```diff
@@ -0,0 +1,427 @@
...
+
+### Use of in-band Tool Call ID
+
+A common workaround pattern is to use a session identifier within CallToolRequests to simulate sessions. This is often model controlled, with the session identifier being reproduced by the LLM. 
+
+In practice, MCP Sessions may be used for other state control - for example availability of Tools, Prompts or Resources - therefore including the sessionId as part of the request/response cycle managed by the Host is the right choice.
+
+### Use of a single `state` value rather than KV store.
+
+A single opaque "state" value mirrors the MRTR design, reduces the chance of KV merge errors, and keeps client behaviour simple (simply echo bytes back).
```
markdroth, line 543: I think there are actually a lot more use-cases here than there are for MRTR request state.

In MRTR, the request state is used solely in the ephemeral workflow and solely for passing data between retries of the same logical request.  This means that there probably won't be too many different values to retain, and all values are going to be closely related (i.e., they will all be implemented by the same tool call implementation).  In that case, I think it's feasible to represent everything in a single string.

In contrast, for session state, I can imagine a larger variety of use-cases that might co-exist in the same session.  Some examples:
- Horizontal functionality implemented in a shared library to do things like authentication or user preferences.
- Application-specific state such as a tool call that unlocks other tool calls in the tool call list.
- Transactional state updated iteratively across multiple tool calls and then committed via another tool call.

All of those things are likely to be implemented by different pieces of code in the MCP server, and those different pieces of code may not be aware of each other, which would make it a bit more challenging to store data for all of them in a single string.

To be clear, I am not necessarily opposed to using a single string here, and I do recognize that it makes the protocol simpler.  But I do want to make sure that we're considering the differences between session state and MRTR request state here and making an informed decision, rather than just blindly copying what we're proposing for MRTR.  Let's be confident that this will actually be sufficient before we go forward.

proposals/0000-data-layer-sessions.md
```diff
@@ -0,0 +1,427 @@
...
+ - Not Required - the SDK will not enforce sessions for MCP operations.
+ - Managed - the MCP Server Author will handle the allocation of sessions (similar to existing Typescript SDK)
+
+ Clients can use the following patterns, and discover whether Sessions are required by making a call or probing the proposed `/discover` endpoint:
+ - Single - the Client will provide a single Session for the connection, or no session if not required. This is similar to existing behaviour.
+ - Managed - the Client will provide an explicit `session.create` operation to return a token, an interface for storage and a reusable token for Client management. 
+
+### resourceAllocation Tool Hint
+
+**For discussion** - it may make sense to include a tool hint to indicate whether or not a Session is associated with expensive resources, hinting that deletion is preferred at the end of the immediate User interaction session.
```
markdroth: I can see the usefulness of this annotation, in the sense that it would tell the client that it needs to create a session before trying to call the tool.  But it's not clear to me that the annotation would really help with deletion, because session data may be useful for a lot more than a single tool call.

markdroth (commented):
Thanks for writing this up, Shawn!  I'm really looking forward to seeing this go through.

proposals/0000-data-layer-sessions.md
evalstate: Added: I've split the second part in to the Backwards Compatibility section.

proposals/0000-data-layer-sessions.md
evalstate: From the current 1442:

> If no session is provided, it’s assumed the server is using them for unassociated or unsolicited requests.

If we have removed dissociated Sampling and Elicitation, this leaves us with notifications only.

Will add some draft prose here to get started.

proposals/0000-data-layer-sessions.md
evalstate: I'll add a note, but for lightweight cases Session persistence could live _entirely_ on the Client.

proposals/0000-data-layer-sessions.md
evalstate: The Server could choose a policy that any session id is acceptable... the idea here was to fail early if invalid sessionids were supplied.

proposals/0000-data-layer-sessions.md
evalstate: Much better, have added comment on possibility of Client error in saving state too.

proposals/0000-data-layer-sessions.md
evalstate: No, but it could help with the case where the Client wants to create a session per tool call.

proposals/0000-data-layer-sessions.md
evalstate: I did think about this a bit, and came to the potentially controversial conclusion that:
1) Exposing a KV store is setting higher expectations than intended of what this should be used for, and invites potential misuse of storage (e.g. sensitive data).
2) We _should_ set an upper size bound for this data (2/4kb), again reinforcing that it is a lightweight convenience mechanism for simple applications that want to attach simple state to the client that gets transmitted with every request.
3) To avoid complexity of message sequencing etc. we want to make sure that complex requirements _are_ handled by the MCP Server author at the Server when necessary.

proposals/0000-data-layer-sessions.md
evalstate: Leaving this TBC (and will keep the current convention in the drafts so it's a one-time change later)!

proposals/0000-data-layer-sessions.md
evalstate: I've modified removed it, and modified the wording it should only be sent on update.

proposals/0000-data-layer-sessions.md
evalstate: Updated link to session hijacking
https://modelcontextprotocol.io/docs/tutorials/security/security_best_practices#session-hijacking

proposals/0000-data-layer-sessions.md
markdroth: The wording here, "unknown sessions", implies (at least to me) that the session ID is not one that was assigned by the server.  If the session was assigned by the server, it wouldn't be unknown.  So this requirement would therefore mean that a server cannot actually choose a policy of accepting any arbitrary session ID, even if the call doesn't actually require a session.

I'm okay with this either way, but we should be clear about exactly which stance we're taking here.

proposals/0000-data-layer-sessions.md
markdroth: I think restricting the size may cause problems for some use-cases, and it's not clear to me that there's any benefit in doing so.  If someone has a use-case that needs a slightly larger amount of data, it seems unnecessarily restrictive to say that we're imposing an arbitary limit that prevents that case from working.  I don't believe there's any size limit in HTTP cookies, as an example.  (There's probably a header size limit in HTTP that limits the size of a single cookie, but I don't believe there's a limit to the total number of cookies that can be set, so there's no limit to the total amount of data.)

Stepping back a bit, I think we really need to write down a set of specific use-cases that we and do not intend to address via session state, so that we can evaluate any specific proposal against those use cases.  In the absence of these use-cases to use as a rubric, it's hard to evaluate whether or not any specific proposal is the right thing.

proposals/0000-data-layer-sessions.md
markdroth: I think this is another situation where we really need to step back and decide which use-cases are or are not in scope for this design.  The example you cited of the client wanting to create a session per tool call seems incompatible with the example I cited earlier of a server wanting to retain state across tool calls.  If the client is assuming one thing and the server is assuming the other, things won't work right.  So I think we need to be clear about the expectations for how sessions can actually be used.

It looks like you removed this tool annotation, which I didn't intend to suggest -- I was really just commenting on the specific example of deletion, which seems unrelated.  But let's first identify what use-cases we are targetting and then decide whether or not this annotation makes sense in the context of those use-cases.

proposals/0000-data-layer-sessions.md
```diff
@@ -0,0 +1,557 @@
...
+
+MCP Sessions are currently either implicit (STDIO), or constructed as a side effect of the transport connection (Streamable HTTP). 
+
+Migrating sessions to the MCP data layer allows MCP applications to handle sessions as part of their domain logic, decoupled from the transport layer. This enables predictable session semantics, especially for Hosts that handle multiple "threads" of context within the application. The ability for MCP Servers to allocate resources on a per-session basis, or be able to provide rich functionality without server-side storage makes MCP suitable for increasingly sophisticated and scaled deployments. 
+
+## Specification
+
+### User Interaction Model
+
+Sessions are designed to be **application-driven**, with host applications determining how and when to establish sessions based on their need.
```
mikekistler, line 65: Not just how and when to establish, but also what subsequent requests are bound to the session.

proposals/0000-data-layer-sessions.md
```diff
@@ -0,0 +1,557 @@
...
+
+It is normally expected that applications will establish one session per conversation thread or task, but this is not required.
```
mikekistler: Does "task" refer to MCP Tasks or the general notion of "task"? Might want to clarify.

proposals/0000-data-layer-sessions.md
```diff
@@ -0,0 +1,557 @@
...
+      "expiresAt": "2026-02-27T15:30:00Z",
+      "state": "bGFuZ3VhZ2U9ZW4="
+    },
+  }
+}
+```
+
+The Client **MUST NOT** send `io.modelcontextprotocol/session` with the sessions/create request.
+
+The Client **MUST** securely associate retained sessions with the issuing Server. The Client will typically establish identity through a mixture of _connection target_ and _user identity_.
```
mikekistler: At least the "illusion" of the issuing Server, which could be a cluster of server instances behind a load balancer.

proposals/0000-data-layer-sessions.md
```diff
@@ -0,0 +1,557 @@
...
+
+That's application-layer guidance that conflicts with what the transport already provides — Streamable HTTP's SSE supports `Last-Event-ID` for resumption, so the transport handles reconnection and missed-event recovery. We shouldn't be re-specifying that here.
+
+Here's the updated section with all three points addressed:
+
+---
+
+#### Receiving Notifications
+
+Clients subscribe to server-initiated notifications using `messages/listen`. This opens a persistent delivery channel (SSE stream over HTTP/logical subscription over STDIO).
```
mikekistler: I cringe when I see that we are baking in a dependency on SSE here. I think we should at least offer an alternative that does not rely on SSE. E.g., allow `messages/listen` to return an array of notifications and complete. Then the client would issue a new `messages/listen` to receive subsequent notifications.

mikekistler: Does this mean that the existing notification channels are deprecated? Maybe that is part of SEP-1442? (Sorry my memory is fuzzy ... if that's the case it would be helpful to state that here)

proposals/0000-data-layer-sessions.md
```diff
@@ -0,0 +1,557 @@
...
+  "jsonrpc": "2.0",
+  "method": "notifications/resources/updated",
+  "params": {
+    "uri": "file:///project/config.yaml"
+  }
+}
+```
+
+
+Global listeners receive **only** broadcast notifications that are not scoped to any session — for example, `notifications/tools/list_changed`. Session-scoped notifications such as resource updates are **never** delivered on a global listener.
```
mikekistler, line 273: This wording suggests that tools/list_changed notifications are always outside a session and resource updated notifications are always within a session. I don't think that's true, based on my understanding of the proposal, so this is probably just a wording thing, but would be good to clarify.

proposals/0000-data-layer-sessions.md
```diff
@@ -0,0 +1,557 @@
...
+1. The Client creates a session (`sessions/create`).
+2. The Client opens a listener (`messages/listen` scoped to that session).
+3. The Client subscribes to a resource (`resources/subscribe` with the session in `_meta`).
+4. When the resource changes, the Server sends `notifications/resources/updated` on the session's listen stream.
+
+Reconnection and missed-event recovery for the listen stream are handled at the **transport layer** (e.g., SSE `Last-Event-ID` for Streamable HTTP).
+
+##### STDIO Transport Behaviour
+
+For STDIO, `messages/listen` acts as a capabilities check. The Client sends the request; the Server responds with `notifications/messages/listen`. The Server **MAY** then send notifications for the declared scope at any time for the duration of the STDIO connection, as it does today.
```
mikekistler: I think this needs elaboration.  Does this mean that NO server-to-client notifications flow in STDIO until a `messages/listen` is sent? If so, that is not "as it is today".

Also, what "capability" is being checked here? The server already declares if it will send tools/list_changed or resouce/updated notifications in its capabilities, which I assume are available from the "discover" request. Is this just "double checking"? Or is it saying something different?

proposals/0000-data-layer-sessions.md
```diff
@@ -0,0 +1,557 @@
...
+1. The Server **SHOULD** implement a policy to remove stale Server maintained session state.
+
+### Data Types
+
+#### SessionMetadata
+
+_The following notes on sessionId are taken from the existing Streamable HTTP Transport guidance._
+
+**sessionId:**
+1. The `sessionId` **SHOULD** be globally unique and cryptographically secure (e.g., a securely generated UUID, a JWT, or a cryptographic hash).
```
mikekistler, line 383: Given that the client must send _both_ the sessionId and sessionData in requests, I don't think it's strictly necessary for sessionIds to be cryptographically secure. The server could easily store a signature for the sessionId in the encrypted sessionData and verify it on subsequent requests.

proposals/0000-data-layer-sessions.md
```diff
@@ -0,0 +1,557 @@
...
+    };
+  };
+}
+```
+
+## Other Work
+
+### SEP-2243 HTTP Standardization
+
+When _meta["io.modelcontextprotocol/session"] is present and using the StreamableHttp transport, the sessionId should be included as `Mcp-Session-Id` in the HTTP Headers.
```
mikekistler, line 516: This again may be something already in SEP-1442, but are transport-level sessions being eliminated? Would be good to state that here.

mikekistler (commented):
Looks good. I left some comments mainly on details to clarify or that might make good discussion topics in the workgroup meeting.

proposals/0000-data-layer-sessions.md
```diff
@@ -0,0 +1,557 @@
...
+```
+
+
+Global listeners receive **only** broadcast notifications that are not scoped to any session — for example, `notifications/tools/list_changed`. Session-scoped notifications such as resource updates are **never** delivered on a global listener.
+
+##### Interaction with Resource Subscriptions
+
+`messages/listen` is the **delivery channel**; `resources/subscribe` is the **subscription mechanism**.
+
+Resource subscriptions **MAY** be associated with a session.  If the `resources/subscribe` request includes session metadata, the resulting notifications/resources/updated notifications are delivered on that session's listen stream. If no session is specified, notifications are delivered on a global listener.
```
pja-ant, line 279: Didn't we say in the summit that we were going to change subscriptions to be self-contained long-lived requests? i.e. the listen would contain the subscriptions, and you'd update subscriptions by re-issuing the listen.

Not sure if we've moved away from that or not.

proposals/0000-data-layer-sessions.md
```diff
@@ -0,0 +1,557 @@
...
+
+HTTP cookies (RFC 6265) provide an existing stateless session mechanism with automatic client-side storage and per-request transmission. The header pattern is well-understood and battle-tested. 
+ 
+This proposal adopts a similar pattern (server-issued opaque tokens that clients return unmodified) but implements it in the JSON-RPC message layer rather than HTTP Headers, enabling consistent session semantics across non-HTTP transports.
+
+### Session Update Sequencing
+
+Because MCP clients may execute multiple requests concurrently (e.g., parallel tool calling), there are no inherent ordering guarantees for request/response cycles. This creates a Last-Write-Wins race condition if the server relies entirely on the client-echoed state token to manage highly mutable data. Additionally Client state saving error conditions are not known to the Server.
+
+To resolve this, servers dealing with concurrent mutations SHOULD NOT rely on the state token. Instead, the server SHOULD use the sessionId as a lookup key for a server-side state management mechanism (e.g., a database, cache, or in-memory store).
```
pja-ant, line 530: Hmm, if there's mutable state in `state` and we're using LWW then it seems this would be rarely useful? We can't rule out clients doing concurrent tool calls, so servers need to deal with that, and it seems you can't really do that in general, so you are forced into adopting LWW semantics, which seems to be of little utility.

What % of stateful servers do we think would want to use `state` and can handle concurrent mutating tool calls? Thinking maybe it would be better to just start with `sessionId` and maybe introduce this later as an optimization if we find a lot of servers would benefit from it?

My sense is that either (a) barely anyone would use this, or (b) people would use it and be prone to race conditions if clients start doing concurrent calls.

proposals/0000-data-layer-sessions.md
mikekistler: How will the STDIO client know which session a notification pertains to? Is the sessionId sent in the notification?

proposals/0000-data-layer-sessions.md
```diff
@@ -0,0 +1,557 @@
...
+## Specification
+
+### User Interaction Model
+
+Sessions are designed to be **application-driven**, with host applications determining how and when to establish sessions based on their need.
+
+It is normally expected that applications will establish one session per conversation thread or task, but this is not required. 
+
+MCP Servers may wish to offer capabilities in a mixture of authentication and session modalities. 
+
```
pja-ant, line 70: We should really define what sessions are. From the rest of the proposal, it *seems* like these things are associated with a session:

1. Resource subscriptions
2. Arbitrary application state

Is the list of tools/resources etc. scoped to a session? (are they allowed to be?)

I'm still somewhat unsure (or at least unconvinced) that we even _need_ sessions...

proposals/0000-data-layer-sessions.md
```diff
@@ -0,0 +1,557 @@
+# Data-Layer Sessions for MCP
+
+> **Status:** Early Draft  
+> **Date:** 2026-02-23  
+> **Track:** transport-wg/sessions  
+> **Author(s):** Shaun Smith
+
+## Abstract
+
+This proposal introduces application level sessions within the MCP Data Layer. Sessions are created by the Client, and allow the Server to store an opaque state token.
```
gjz22: Minor: the wording "allow the Server to store an opaque state token." is a bit unclear here. Do you mean "and allow the Server to send to the Client an opaque state token that will be sent back for each relevant request"?

proposals/0000-data-layer-sessions.md
evalstate: Thanks @markdroth - that's clear now.

proposals/0000-data-layer-sessions.md
evalstate: Yes, that's much better wording.

proposals/0000-data-layer-sessions.md
evalstate: I do agree.

The spec doesn't make any guarantee on notifications being delivered - although some SDKs support something _like_ this as part of SSE resumption. I don't think we should make stronger promises here than in the transport spec itself. 

I think we should cover the _scope_ of what you would get on the 1442 endpoint here, but not the mechanism.

proposals/0000-data-layer-sessions.md
evalstate: I understand ths and my view is.
1) I'd rather not have it than impose ordering/delivery guarantees
2) There is a class of (fun), lightweight application that it enables. For example a Zen Koan tool which recites a phrase and increments a client-side counter, ready for the next ask. (Local STDIO servers which want to store simple location/project information etc. would benefit too).  I've left fun in brackets this feels... very hackable in a good way.

proposals/0000-data-layer-sessions.md
```diff
@@ -0,0 +1,586 @@
...
+> An MCP “session” consists of logically related interactions between a client and a server, beginning with the initialization phase. To support servers which want to establish stateful sessions:
+
+Sessions allow Clients and Servers to maintain shared contextual state across sequences of Tool Calls 
+ - Maintain shared contextual state between Tool Calls.
+ - Rehydrate application state on resumption.
+ - Servers may choose to customise MCP Server features based on the presence or absence of a Session.
+ - Servers may scope Notifications to specific sessions. 
+
+Sessions are _not_ intended to provide a guarantee of MCP Protocol State. For example:
+ - Tools Lists
```
pja-ant: To be clear: are we explicitly saying here that the lists of tools are not scoped to a session? So you cannot do things like `enable_admin_tools()` within a session to enable new tools only for this session? (I'm fine with that, just want to be explicit about it). I'm not sure what "provide a guarantee of MCP protocol state" means in practice.

pja-ant: May need to do some due diligence here to understand how many (if any) people are using sessions like this today.

proposals/0000-data-layer-sessions.md
evalstate: I'll update this to include session on the notification.

proposals/0000-data-layer-sessions.md
evalstate: What I want to convey is that the server should assess requests on a case-by-case basis, and that the Client only ever has a cache of the state at some point in the past. If we think of `unlock admin tools` after an elicitation, then we would need to refresh the client side cache. We'd usually do that with a notification. The Server may wish to record that against a session, or it may wish to refuse a tool call after an elapsed period of time (say 30 minute cooldown). I don't think making guarantees is flexible enough.

proposals/0000-data-layer-sessions.md
evalstate: do you think it's worth adding "logical issuing Server". not sure how best to phrase/incorporate that.

proposals/0000-data-layer-sessions.md
evalstate: I've added some more commentary on scoping

proposals/0000-data-layer-sessions.md
```diff
@@ -0,0 +1,591 @@
...
+
+With the removal of the `initalization` phase and associated lifecycle semantics, data-lyer sessions are scoped as follows:
+
+Sessions allow Clients and Servers to bind a sequence of MCP requests into an application-defined context recognized by the Server. Sessions provide contextual association, not snapshot semantics. A session can scope:
+ - processing state across multiple operations;
+ - server-managed resources or allocations associated with that context;
+ - subscriptions and delivery of server-initiated messages related to that context; and
+
+Sessions may influence how the Server evaluates requests and responses, but it does not provide a guarantee of MCP Protocol State, including:
+- Tool Lists
```
evalstate, line 37: (reminder) I think we need to look at how SEP-1442 handles dynamic lists/cache invalidation.

evalstate:
> Can we consider giving a way for servers to signal the requirement to use sessions?

@ggoodman -- `SESSION_REQUIRED` added - either fail fast or fail on SEP-1442 `discover` call if needed. I think that covers it?

proposals/0000-data-layer-sessions.md
evalstate: > To be clear: are we explicitly saying here that the lists of tools are not scoped to a session?

I have tried to improve the wording here; the session doesn't "pin" the state of the tool list ; Tool/Prompt Lists are available per-request and can be distinguished by Session. But the Session doesn't guarantee the stability of that list (any more than it is stable today).

proposals/0000-data-layer-sessions.md
evalstate: The optional `GET` handler is deprecated, meaning that global notifications (e.g. those not directly associated with a request) need a comms. channel - so it's replaced by `messages/listen` . From SEP-1442:


> This is achieved using a **dedicated messages/listen RPC. For the HTTP transport, the client sends this request via POST, and the server's response is an open SSE stream, with a MessagesListenNotification sent as the first event. For the STDIO transport, this RPC is used for a simple request/response capabilities check. This RPC replaces the existing /GET endpoint behavior for Streamable HTTP today.
> Client-initiated streaming MAY be associated with a session. If no session is provided, it’s assumed the server is using them for unassociated or unsolicited requests.

proposals/0000-data-layer-sessions.md
```diff
@@ -0,0 +1,633 @@
...
+
+`state` **MUST** be retained by the Client and sent with future requests for that session.
+
+#### Using Sessions
+
+To use a Session the Client request includes SessionMetadata in `_meta["io.modelcontextprotocol/session"]`:
+
+1. The Server MUST treat that sessionId as the session context for processing the request.
+1. The `sessionId` in the response MUST exactly match the sessionId from the request.
+1. For operations that require a Session, if the request does not include `_meta["io.modelcontextprotocol/session"]`, the Server **MUST** reject the request with `SESSION_REQUIRED`.
```
Randgalt, line 131: Why not make this the defined way to create a session? We could save a round trip by removing the `sessions/create` message. Any time the server requires a session it can return the `SESSION_REQUIRED` response. JSON-RPC error responses are allowed to have data so the error could carry the session creation details.

proposals/0000-data-layer-sessions.md
```diff
@@ -0,0 +1,633 @@
...
+}
+```
+
+> For testing purposes MCP Clients that support sessions declare an `experimental/sessions` capability to simplify testing.
+
+### Protocol Messages
+
+#### Creating Sessions
+
+Clients begin a session with an MCP Server by calling `sessions/create`.
```
Randgalt, line 89: It's a shame that we're able to get rid of `initialize` message only to now have a replacement.

proposals/0000-data-layer-sessions.md
Randgalt: Or, maybe, the session creation can be incorporated as a `_meta` response to any call. So, that if a new session is required it becomes an attachment to any response (something like COOKIE headers in HTTP).

proposals/0000-data-layer-sessions.md
gjz22: Agreed. It would mean just for sessions there would need to be an initial handshake before using a server. Thoughts on how to do this without a `sessions/create`:
1. Try then fail: Try a method that requires a session but don't pass a session id. Server method fails with a SessionRequiredError. Client goes through the session handshake with the Server. Client retries the original method now passing the session id.
2. Client creates the session id as something unique to that client. The server just accepts the client's session id, attaching data to it.

Nice thing about this is that it removes the need for capability negotiation as well.

proposals/0000-data-layer-sessions.md
gjz22: What about a list of strings? Responses return a list `stateActions` each with properties `action` (remove|add|clear) and `value` and the entire list is passed to the server on each request.

This deals with the parallel tool call issue as state actions can be handled deterministically (removes before adds for example).

proposals/0000-data-layer-sessions.md
```diff
@@ -0,0 +1,633 @@
...
+
+## Use Cases and Scope
+
+### Scope
+
+The current Transport specification defines sessions as follows: 
+
+> An MCP “session” consists of logically related interactions between a client and a server, beginning with the initialization phase.
+
+With the removal of the `initalization` phase and associated lifecycle semantics, data-lyer sessions are scoped as follows:
```
Randgalt, line 29: fix: `data layer`

proposals/0000-data-layer-sessions.md
```diff
@@ -0,0 +1,633 @@
...
+```
+
+
+Global listeners receive **only** broadcast notifications that are not scoped to any session — for example, `notifications/tools/list_changed`. Session-scoped notifications such as resource updates are **never** delivered on a global listener.
+
+##### Interaction with Resource Subscriptions
+
+`messages/listen` is the **delivery channel**; `resources/subscribe` is the **subscription mechanism**.
+
+Resource subscriptions **MAY** be associated with a session.  If the `resources/subscribe` request includes session metadata, the resulting notifications/resources/updated notifications are delivered on that session's listen stream. If no session is specified, notifications are delivered on a global listener.
```
Randgalt, line 279: Can there be more than one "session listen stream"? It would be nice to clarify this to say that for a given session there can be exactly one listen stream.

proposals/0000-data-layer-sessions.md
Randgalt: So, this would be `resources/subscribe` request which then expects an SSE stream that returns `notifications/resources/updated` events? I assume the implication is that whatever version of a the subscribed to resource is at at the point of `resources/subscribe` becomes the basis for update notifications.

proposals/0000-data-layer-sessions.md
```diff
@@ -0,0 +1,633 @@
...
+Servers SHOULD define a lifecycle policy for subscriptions — for example, scoping them to a session if one exists, or applying a TTL for sessionless subscriptions.
+
+The typical flow:
+
+1. The Client creates a session (`sessions/create`).
+2. The Client opens a listener (`messages/listen` scoped to that session).
+3. The Client subscribes to a resource (`resources/subscribe` with the session in `_meta`).
+4. When the resource changes, the Server sends `notifications/resources/updated` on the session's listen stream.
+
+Reconnection and missed-event recovery for the listen stream are handled at the **transport layer** (e.g., SSE `Last-Event-ID` for Streamable HTTP).
```
Randgalt, line 290: Servers should be able to terminate the stream at any time (e.g. after an internal TTL). This should be expected behavior by the client and the client can reconnect if it chooses (and use `Last-Event-ID`).

jeffyaw:
I've been building an MCP proxy layer that does session affinity by routing `Mcp-Session-Id` to pinned backends. A couple of things from that experience that might be relevant:

1. **Proxy/infrastructure visibility** Moving session identity into the JSON body (`_meta`) means intermediary layers (proxies, load balancers, API gateways) lose the ability to route based on session without parsing the body. The current `Mcp-Session-Id` header is very routable. Is the intent that the transport-level header (per SEP-2243) would still be the primary mechanism for infrastructure routing, with `_meta` being the data-layer complement?

2. **Session creation latency** For proxy-hosted servers where the backend may be cold-started or horizontally scaled, `sessions/create` as a separate RPC adds a round trip before any useful work happens. The suggestion from @Randgalt about folding session creation into the first response (cookie-style) resonates with what I've seen in practice.

kurtisvg (approved):

kurtisvg:
Closing this per #27

PR #31, opened 2026-04-13
SEP-XXXX: Required Capabilities for Tools

gjz22:
This is a draft SEP addressing cross-request capability coordination for the per-request client capabilities introduced by [SEP-2575](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/2575).

## Motivation and Context
SEP-2575 moves client capabilities from a single initialization handshake to per-request declarations in `_meta`, but some requests (notably `tools/list`) need to know what capabilities will be available on other requests (notably `tools/call`). When the mismatch goes undetected until call time, it surfaces in two damaging ways: raw capability errors propagate to end users who know nothing about MCP, and the model wastes tokens planning multi-step work toward a tool that can never run.

This SEP proposes that servers annotate each tool with the capabilities it hard-requires, via an `io.modelcontextprotocol/requiredCapabilities` field in the tool's `_meta`. The annotation has the same shape as `ClientCapabilities`, so it covers extension capabilities (such as tasks) with no additional mechanism. Clients match their own capabilities against the annotations and decide how to respond: filter the tool out, present it as unavailable, or allow it and fail with a clear error. Tools that can operate in **hybrid mode** — using a capability when the `tools/call` request declares it and falling back otherwise — need no annotation. Servers still validate at call time per the core spec (`MissingRequiredClientCapabilityError`, `-32021`), and filtering clients treat such failures as a signal to refresh `tools/list`.

The SEP discusses the server-side vs. client-side filtering trade-off in detail: client-side keeps `tools/list` responses identical for every client (simplifying caching, and MCP proxy servers in particular) and leaves clients flexibility in how to respond, while server-side allows more complex filtering logic and keeps capability filtering alongside the user-based filtering servers already do.

Alternatives are discussed in detail:
1. Related request capabilities (`relatedRequestCapabilities` keyed by method name in `_meta`, enabling server-side filtering)
2. No capability coordination (all tools returned regardless, failures at call time)

Modifications considered but not included for simplicity: a top-level `requiredCapabilities` tool field instead of `_meta`, and a `whenCapabilitiesUnavailable` fallback-tool annotation for static substitution.

See SEP for details.

## How Has This Been Tested?
Not tested yet, that will happen when we implement.

## Breaking Changes
This proposal is additive to SEP-2575 and does not introduce breaking changes on its own. Clients and servers that do not understand the annotation see today's behavior. See SEP for details.

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context
N/A


proposals/XXXX-capability-scope.md
```diff
@@ -0,0 +1,782 @@
...
+
+## Abstract
+
+[SEP-1442](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1442)
+introduces per-request client capabilities, where the client declares
+its supported capabilities in the `_meta` field of each request rather
+than negotiating them once during initialization.  However, that
+proposal does not specify the *scope* of those capabilities: whether a
+capability declared on one request should be assumed to hold for other
+requests, or whether each request's capabilities are independent.
```
pja-ant: A bit of a semantical nit, but I'm not sure we should use "scope" here. It's really not about scope and I find the "globally scoped" phrasing confusing (what does "global" mean?). With 1442 we have a stateless protocol, so the only scope that exists from MCP's perspective is the request.

What is being discussed is if/how servers should factor in capabilities sent on `list/tools`, which I think this  SEP does a good job of exploring. I'd recommend dropping the "scope" verbiage.

proposals/XXXX-capability-scope.md
```diff
@@ -0,0 +1,782 @@
...
+what capabilities will be available at `tools/call` time.
+
+## Specification
+
+### Global Capability Scope
+
+When a client sends per-request capabilities (as defined in SEP-1442),
+the following rules apply:
+
+1. A client **SHOULD** send the same `clientCapabilities` object on every
```
pja-ant: This is where the scoping ("every") makes things confusing. What does "every" request mean? Are clients not allowed to change capabilities?

A more direct phrasing would be something like "A client should not assume that a tools list returned with one set of supplied capabilities is the same that another set of capabilities would return" -- with the implication being that if a client changes capabilities then it should probably re-fetch tools. It should also not cache tools across capability sets.

proposals/XXXX-capability-scope.md
```diff
@@ -0,0 +1,782 @@
...
+   request.  The capabilities declared represent the client's overall
+   capability set, not capabilities specific to the request type.
+
+2. A server that receives capabilities on any request **SHOULD** assume
+   that those capabilities apply to all requests from that client.
+   Specifically, if a server receives capabilities on a `tools/list`
+   request, it **SHOULD** use those capabilities to determine which tools
+   to return, assuming the same capabilities will be present on
+   subsequent `tools/call` requests.
```
pja-ant: It should not assume that capabilities apply to subsequent requests: every request is independent. The capabilities may change on subsequent `tools/call` requests and the server will have to handle that (e.g. by rejecting the request).

proposals/XXXX-capability-scope.md
```diff
@@ -0,0 +1,782 @@
...
+
+3. If a client's capabilities change (e.g., a user enables or disables
+   a feature), the client **SHOULD** re-issue applicable list requests
+   (e.g., `tools/list`, `resources/list`) with the updated capabilities
+   rather than relying on cached results.  This ensures the server can
+   return results that reflect the new capability set.
+
+4. A server **MUST NOT** assume capabilities are available if the client
+   has not declared them on any request.  The absence of capabilities
+   is always a safe default.
```
pja-ant: These are good. I do think we're on the same page, but the language needs to be phrased in a more stateless way where every request is treated in isolation (since you have no other choice).

proposals/XXXX-capability-scope.md
```diff
@@ -0,0 +1,782 @@
...
+   capability sub-fields.
+
+3. **Consistency**: The tool list returned by `tools/list` always
+   matches what the client can actually execute via `tools/call`.
+
+4. **Compatibility with stateless design**: In a stateless server, each
+   request carries the full capability set, so any server instance can
+   make the correct decision without shared state.
+
+### Alternatives Considered
```
pja-ant: Thanks for including these in so much detail. I think it could be very useful to provide a comparison matrix across all options to give a high-level overview.

pja-ant (commented):
This is great, thanks for putting it together! 

Added some comments, mostly semantical things that would make it a bit more precise (in my opinion), but high-level I think this is the right framing of options. Would love to see a comparison matrix to help us debate trade-offs.

FWIW, I'm really not sure what the right thing is here. I was originally leaning towards your proposed option but can see pros/cons versus others as well.

proposals/XXXX-capability-scope.md
```diff
@@ -0,0 +1,782 @@
...
+- **Status**: Draft
+- **Type**: Standards Track
+- **Created**: 2026-04-11
+- **Author(s)**: Gabriel Zimmerman (@gjz22)
+- **Sponsor**: None
+- **PR**: https://github.com/modelcontextprotocol/transports-wg/pull/{NUMBER}
+
+## Abstract
+
+[SEP-1442](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1442)
```
ggoodman:
```suggestion
[SEP-2575](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/2575)
```

proposals/XXXX-capability-scope.md
kurtisvg: More broadly, I feel like we need to define what criteria should _can_ be used to affect the outcome of things like tools/list, resources/list. The wording in SEP-2567 says:

> `tools/list`, `resources/list`, and `prompts/list`: there is no longer a per-session or per-connection scope for their results to depend on. Lists can still change for other reasons — a server deploys a new version, a user's plan or granted scopes change — and this SEP does not enumerate or restrict those.

It seems like we need to clearly enumerate in which case the tools/list should be cached or shared. In what instances should clients refresh the list?

proposals/XXXX-capability-scope.md
```diff
@@ -0,0 +1,782 @@
...
+introduces per-request client capabilities, where the client declares
+its supported capabilities in the `_meta` field of each request rather
+than negotiating them once during initialization.  However, that
+proposal does not specify the *scope* of those capabilities: whether a
+capability declared on one request should be assumed to hold for other
+requests, or whether each request's capabilities are independent.
+
+This proposal specifies that per-request client capabilities SHOULD be
```
kurtisvg: I'm not a big fan of this wording -- it feels like we should be giving the client guidance on when it should refresh tools/list, not what the server should assume. The client could change capabilities at any time, and it should be the clients responsibility to refresh the given tools/list accordingly.

kurtisvg (commented):
At it's core, I think this issue should be reframed as "what information can be used to filter */list calls", and probably should explicitly call out auth/user identity in addition to capabilities. 

It also needs to be refocused on defining client behavior more sot than server behavior, because in a stateless world the server doesn't differentiate between different clients, so it's up to the client to refetched cached info (and not the server to try and enforce that).

proposals/XXXX-capability-scope.md
gjz22: Global here means what it means today in the protocol. That you have one set of capabilities and those capabilities apply globally to all requests as opposed to being able to specify per request what capabilities the client has for that request only.

proposals/XXXX-capability-scope.md
gjz22: Makes sense.

proposals/XXXX-capability-scope.md
gjz22: Wording aside, to be clear, the assumption here is that if a client declares capabilities on tools/list, unless the client changes capabilities (in which the client should refetch tools), the server should assume the client will have the same capabilities on tools/call

proposals/XXXX-capability-scope.md
gjz22: Added (quickly but will revise). I included only the main option, and alternative 1 and alternative 2 because I consider these the main viable options.

ZachGerman:
Firstly, I want to call out that I believe there's a significant rift in desire here from closed-system and public server perspectives.

Since I've put a lot of thought into Alt 1 (e.g. [[1]](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2487),[[2]](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1385)), I'd like to weigh in on the differences between this and Alt 1, which I believe will have significant impact on this protocol going forward:

**Server-side filtering** (vs Client-side): (assuming no room for filtering parameters on `tools/list` requests) Either the servers must disclose their entire tool list, or the client must disclose their entire capability set. Servers today exist to expose tools for calling. Clients today don't exist to expose their capability set, and shouldn't be required to outside of closed-systems to achieve full protocol functionality (capability-specific CVEs are scary "New MCP sampling exploit takes users by surprise!"). From a closed-system owner perspective, there may be desire to control client behavior via the server's tool filtering. From a public server perspective, they rely on independent clients to be competent in selecting the right tools given the definitions, and may even appreciate the extra exposure provided by the opportunity to convey their entire tool list to a client. Maybe there's room for both. I'm not sure.

**Substitution**: This is filtering.

**Parameterized adaptation**: I think [this](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2487) is a better approach to support this than forcing clients to disclose their full capability details. Again, a closed-system owner may not have the same privacy or security concerns here that client users calling public servers would. Say a user is operating a client with some kind of private capability. They may not want to share that with all the public servers they interact with.

**Simplicity for client implementers**: I think the differences here are trivial compared to some of the problems being solved outside this discussion. I developed matching logic for my implementation example for [#1385](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1385) very quickly. It's definitely worth it in order to maintain client discretion in exposing capabilities.

**Simplicity for server implementers**: I disagree with the label of MEDIUM here for Alt 1. They wrote the tool. They know what's required for it to execute successfully.

**Transparency to the client**: While I agree with the assessment, I think if you're going to have a row for this, you shouldn't leave out the row for the inverse: "Exposure (forcing transparency) of client to the server". It's important to remember that public servers advertise themselves and strive to be transparent in order to earn trust from clients, who have to be careful of what servers they reach out to (just like we're all careful of the websites we visit).

**True per-request capability scoping**: Agree.

------------

Overall, I think there's much to be learned from the evolution of `user-agent` strings in HTTP here. The timeline was something akin to:
1. Full client disclosure of incredibly detailed capabilities
2. Attackers began fingerprinting and exploiting clients
3. `user-agent` verbosity was reduced heavily
4. In came [HTTP Client Hints](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Client_hints)

So, unless we start limiting capability declaration (I worry about custom capabilities before they have their time here) in similar ways, we'd be on course to start from point 1.

proposals/XXXX-capability-scope.md
gjz22: I will revamp the whole SEP (using Alternative 1 as the main option), but for (2) above, I'll weaken the language. Replacing `A server that receives capabilities on any request **SHOULD** assume` with `A server that receives capabilities on any request **MAY** use as guidance` while noting the stateless nature and error handling following SEP-2549 (TTLs).

proposals/XXXX-capability-scope.md
gjz22: Revamped the wording. Since the proposal is to have the client filter, if the client is filtering AND they still get an error, that's a signal that the client should refresh the list of tools to get the new capabilities.

proposals/XXXX-capability-scope.md
gjz22: Whole SEP revamped.

PR #42, opened 2026-06-17
Proposal: i18n via per-request language negotiation

SamMorrowDrums:
This proposal was reviewed by the Transport Working Group.

## Working Group outcome

**Vote:** 3 Accept, 3 Accept with Changes.

The requested changes were:

- Stronger examples showing why language preference cannot be replaced by a tool-call argument.
- Stronger justification for defining this behavior in the core specification rather than as an extension.

Those changes are now incorporated in the synchronized proposal text.

## Continued standards-track discussion

Further discussion and development continue in [modelcontextprotocol/modelcontextprotocol#2792](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2792).

**Reference implementation:** [modelcontextprotocol/typescript-sdk#2158](https://github.com/modelcontextprotocol/typescript-sdk/pull/2158).


proposals/XXXX-i18n-language-negotiation.md
```diff
@@ -0,0 +1,687 @@
...
+### Scope: user-facing content (and beyond)
+
+The mechanism itself is entirely optional (see the note at the start of
+[Specification](#specification)). When a server **does** choose to honor
+`acceptLanguage`, this SEP scopes the negotiation to **user-facing
+content**: strings that an MCP client surfaces directly in a user
+interface and that the human user reads. Typical user-facing fields
+(non-exhaustive) include:
+
+- `title` and `description` on tools, resources, prompts, and server
```
kurtisvg: one of the pieces for feedback was more clarification on what should change. 

Does this include descriptions for parameters in a tools schema for example?

proposals/XXXX-i18n-language-negotiation.md
```diff
@@ -0,0 +1,687 @@
...
+  metadata (rendered in tool pickers, capability lists, settings, etc.)
+- Error `message` strings intended for surfacing in a UI
+- Notification messages (e.g. `logging/message`) that may be rendered to
+  the user
+- Any other field the host explicitly displays to the user without
```
kurtisvg: I think we should try to expictly call out which fields these are, as I don't know that there is concencus.

proposals/XXXX-i18n-language-negotiation.md
SamMorrowDrums: Good question — addressed in `254b291`. The Scope section is rewritten as three explicit buckets (MUST / MAY / MUST NOT translate), anchored to the schema's own framing of each field.

To your specific example: yes, property `description` and `title` inside `Tool.inputSchema` / `outputSchema` are included, but explicitly as **MAY**, with a stated caveat that translating them changes what the model sees and can affect tool-selection and agent behaviour. The Codex implementation (`codex-rs/tools/src/mcp_tool.rs`, `codex-rs/codex-mcp/src/rmcp_client.rs`) wires these straight into the function definition sent to the model, which is the reason for the caveat rather than a blanket MUST.

`Tool.description` itself is in the same MAY bucket for the same reason — the schema describes it as "a hint to the model".

proposals/XXXX-i18n-language-negotiation.md
SamMorrowDrums: Done in `254b291`. The open-ended "any other field the host displays" bullet is gone. The replacement is an enumerated list across three buckets:

- **MUST translate** — fields the schema classifies as display-only (`BaseMetadata.title` on every type, `ToolAnnotations.title`, elicitation messages and form-schema property `title`/`description`, `ProgressNotification.message`, `JSONRPCError.message`).
- **MAY translate, with caveat** — model-facing or dual-purpose fields (`Tool.description` and the other "hint to the model" descriptions, `Prompt`/`Implementation` descriptions, `inputSchema`/`outputSchema` property `title`/`description`, body content from `tools/call` / `resources/read` / `prompts/get` including MCP Apps UI resource bodies, `LoggingMessageNotification.data` where contractually user-facing).
- **MUST NOT translate** — machine-interpreted values (`name`, URIs, MIME types, JSON Schema property keys, enum tokens, capability identifiers, `_meta` keys).

proposals/XXXX-i18n-language-negotiation.md
```diff
@@ -0,0 +1,740 @@
...
+
+## Motivation
+
+MCP exposes user-facing strings on tools, resources, prompts, and server
+metadata (`title`, `description`, `text` content blocks, error messages, etc.)
+that are intended for display in user interfaces. The specification today
+provides no mechanism for a client to request these strings in a particular
+language, or for a server that supports multiple languages to advertise which
+one it returned. This forces ad-hoc solutions and discourages MCP servers
+from investing in i18n at all.
```
kurtisvg: Can we provide a concrete example of where i18n would be useful here? I don't work with models in other langauges, so I'm unclear how they handle different languages today. 

Are there benefits to keeping some stuff as the same language?

proposals/XXXX-i18n-language-negotiation.md
```diff
@@ -0,0 +1,740 @@
...
+translate. The classification follows the schema's own framing of each
+field: where a doc-comment says "intended for UI" or "human-readable
+title", the field is display-only; where it says "hint to the model" or
+"improve the LLM's understanding", the field is model-facing.
+
+#### MUST translate
+
+These fields are classified by the schema as display-only. A server
+honoring `acceptLanguage` **MUST** localize them when a localized form
+is available.
```
kurtisvg: nit: this looks like maybe a "SHOULD" -- as "when a localized form is available seems to be an example where it wouldn't localize it.

proposals/XXXX-i18n-language-negotiation.md
```diff
@@ -0,0 +1,740 @@
...
+   which is itself a conformance hazard. Byte-equality is
+   unambiguous and trivial to verify.
+
+The cost of the byte-match rule is the
+[normalization footgun](#normalization-footgun-and-intermediary-configuration):
+operators using header-rewriting CDN features for per-language caching
+must reconfigure them to leave `Accept-Language` either verbatim or
+absent.
+
+### Why `-32005` rather than `-32001`
```
kurtisvg: There was another PR that changed these error codes to 32020+, we should make sure this aligns with that.

proposals/XXXX-i18n-language-negotiation.md
```diff
@@ -0,0 +1,740 @@
...
+  **SHOULD** fall back to a server-defined default and **MUST NOT** return an
+  error solely because of an unmatched preference.
+
+#### `contentLanguage` (response)
+
+- A server that selected a language in response to `acceptLanguage`, or that
+  is aware of the language of the user-facing content it returned, **MUST**
+  include `io.modelcontextprotocol/contentLanguage` in `result._meta` on a
+  successful response, or in `error.data._meta` on an error response (see
+  [Error responses](#error-responses) below).
```
kurtisvg: I think this should by limited to "selected alanguage in response to acceptLanguage" -- the "if a server is aware" it a bit unclear what that means. I think if it was sent, the server is aware.

kurtisvg (commented):
Overall, supportive of voting on this to bring to the CMs.

proposals/XXXX-i18n-language-negotiation.md
SamMorrowDrums: Addressed in `7b50dd8`. I added concrete public-sector document and mixed-language web-search examples. The new text explains why official or quoted content may need to remain verbatim, why model-facing language can affect behavior, and why the SEP should set a recommended UI-facing floor while leaving returned-content policy to server authors and upstream providers.

proposals/XXXX-i18n-language-negotiation.md
SamMorrowDrums: Agreed and addressed in `7b50dd8`: the display-only bucket is now **SHOULD translate**, including the conditional case where a localized form is available.

proposals/XXXX-i18n-language-negotiation.md
SamMorrowDrums: Updated in `7b50dd8`. `HeaderMismatch` now uses the settled `-32020` allocation throughout, citing merged modelcontextprotocol/modelcontextprotocol#2907 and the current SEP-2243/schema. The provisional `-32005` text and stale allocation references are removed.

proposals/XXXX-i18n-language-negotiation.md
SamMorrowDrums: Agreed and addressed in `7b50dd8`. The requirement is now limited to a server that selected a language in response to `acceptLanguage`; related error-response and conformance language was aligned with that trigger.

kurtisvg:
The Transport WG voted for this today to move forward as an SEP. The vote was:
- 3 Accept
- 3 Accept w/ Changes


Changes requested:
- Better examples of why this functionality was necessary (e.g. can't be replaced by an argument in a tool call)
- Justification on why this functionality should exist in the core spec (instead of being an extension)

SamMorrowDrums:
The Transport WG vote passed: **3 Accept, 3 Accept with Changes**. The requested changes have been incorporated into the synchronized proposal. Standards-track work continues in [modelcontextprotocol/modelcontextprotocol#2792](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2792).

PR #43, opened 2026-06-17
add streaming brief

evalstate:
Add Streaming Tools brief to start working group.


proposals/streaming-brief.md
```diff
@@ -0,0 +1,31 @@
...
+ - Database Queries streaming large volumes of data at variable rates
+ - Computer Use or Sandbox tools emitting output and error streams.
+
+MCP is currently limited to sending updates via `notification/progress`[^1], or using undocumented `resource/subscribe` side-channel approaches to share progress.
+
+This limits the use of MCP for tasks that would benefit from early information sharing e.g. error conditions from delegated tasks, or content generation that does not meet requirements. It also allows Users to interact with the information  
+
+## Key Requirements:
+
+The requirements for MCP Tools are to:
```
pja-ant, line 21: nit: this may go beyond tools (e.g. streaming resources)

proposals/streaming-brief.md
```diff
@@ -0,0 +1,31 @@
...
+ - Provide a common way for MCP Servers to efficiently deliver partial results Hosts.
+ - Signal whether updates are intended to be supervised by Model or User to allow monitoring or other interventions (steer/cancel) during the execution. 
+ - Handle supported modalities and data types cleanly (e.g. Text, Image, Audio, Structured) 
+ - Consider conventions for handover and connectivity to live streaming models (realtime voice or video models for example).
```
pja-ant, line 25: I think this is a good list for consideration. I would be ok if we didn't solve *all* of this in an initial streaming solution, e.g. live media streaming will be a big area with unclear model integration (IMO). It does feel important, but if v1 only solved discrete, non-live streaming then I think that would be ok. (I'd actually expect realtime media handling to warrant its own WG).

proposals/streaming-brief.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Key Design Decisions:
```
pja-ant, line 27: I would say that relationship to Tasks is an open question, particularly for the interactive version where the user/model can interact back with the server mid-tool call. Non-task tool calls are typically considered synchronous from a model-integration point of view and having the model address an in-progress tool call would be quite a deviation in how tools work.

kurtisvg: +1

proposals/streaming-brief.md
```diff
@@ -0,0 +1,31 @@
...
+
+ - When to use MCP Apps vs. new built-ins for delivering interactive content.
```
kurtisvg, line 29: I also think we need to consider how this works with partial results where the result of the results might not be warrented. 

e.g. if I return 10k results from the table, I probably want the results streamed. Should pagination be handled by the tool protocol or the tool API?

evalstate, line 29: Need to make sure that the design goals/constraints work well together with Tasks - making sure that it is clear to Users of the Protocol when/how Tool Streaming would interact with Tasks.

evalstate, line 29: @kurtisvg -- need to be clear on working with the limitations of JSON-RPC.

evalstate, line 29: @markdroth -- understand how this works with existing event streams, ordering, need for bidirectional comms.

proposals/streaming-brief.md
```diff
@@ -0,0 +1,31 @@
...
+MCP is currently limited to sending updates via `notification/progress`[^1], or using undocumented `resource/subscribe` side-channel approaches to share progress.
+
+This limits the use of MCP for tasks that would benefit from early information sharing e.g. error conditions from delegated tasks, or content generation that does not meet requirements. It also allows Users to interact with the information  
+
+## Key Requirements:
+
+The requirements for MCP Tools are to:
+ - Provide a common way for MCP Servers to efficiently deliver partial results Hosts.
+ - Signal whether updates are intended to be supervised by Model or User to allow monitoring or other interventions (steer/cancel) during the execution. 
+ - Handle supported modalities and data types cleanly (e.g. Text, Image, Audio, Structured)
```
evalstate, line 24: Define tools and behaviour **early** --
@evalstate to get agent/multimodal examples, @kurtisvg / team to help with large data xfer.

Issue #36, opened 2026-06-08
Exploring the Future of "Sessions" (Thread IDs / Conversation IDs)


kurtisvg:
## Background & Current State

Recently, the MCP core maintainers made the decision to remove "Sessions" from the protocol. 

Historically, sessions were included in the protocol but suffered from significant ambiguity and implementation complexities. After reviewing the trade-offs between stateful and sessionless architectures, the decision was made to move to a sessionless baseline. 

You can read the full context of this decision here:
* **Transports WG Decision Doc:** [sessions-vs-sessionless-decision.md](https://github.com/modelcontextprotocol/transports-wg/blob/main/docs/sessions-vs-sessionless-decision.md)
* **Core Maintainers Vote & Discussion:** [#2536](https://github.com/modelcontextprotocol/modelcontextprotocol/discussions/2536)

###  Pushback

Since the decision, we've received pushback from the community highlighting scenarios where maintaining state across a series of interactions feels necessary. There is an active debate, including alternative proposals to retain or rework this functionality (see [PR #2822](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2822)).

We are opening this issue to formally collect **concrete use cases** for "sessions", and to discuss what a path forward might look like.

##  Experimental Extension

Before bringing sessions back into the core protocol, we are considering introducing it as an **experimental extension**. However, to do this successfully, we need to resolve the core challenges that caused us to remove sessions in the first place. 

If we reintroduce this concept, we also need to move away from the term "Sessions." The word is heavily overloaded across different networking and application layers. Moving forward, we are proposing terms like **Conversation IDs** or **Thread IDs** to more accurately reflect their purpose: grouping a logical sequence of MCP requests.

### Challenges

For any Thread ID / Conversation ID proposal to succeed, we need to clearly define its scope. Previous implementations struggled because the boundaries were too blurry. We need to answer:

1. **Lifecycle Management:** 
   * When exactly does a Thread ID start and stop? 
   * Is the lifecycle explicitly managed by the client (e.g., an initialization call), or implicitly created on the first tool call?
   * How do we handle cleanup and timeouts for abandoned threads?
2. **Scope of Impact:** 
   * What protocol capabilities should a Thread ID be allowed to mutate? 
   * Does a Thread ID only provide context to `tools/call` execution?
[...]

PR #47, opened 2026-07-14
feat: add problem statement for reintroducing sessions

kurtisvg:
Summarizes the discussion from #36 into some clear questions and use cases.

### DIFF
--- a/docs/sessions-reintroduction-problem-statement.md
+++ b/docs/sessions-reintroduction-problem-statement.md
@@ -1,0 +1,361 @@
+
+# Problem Statement: Reintroducing "Sessions" (Thread / Conversation IDs)
+
+## Background
+
+MCP historically included sessions: an `Mcp-Session-Id` header on Streamable
+HTTP, and an implicit session tied to process lifetime on stdio. In practice the
+concept blurred two distinct jobs: transport-level bookkeeping (protocol
+version, capability negotiation) and application-level state (a conversation
+thread, stateful tool operations). This ambiguity caused problems: wildly
+inconsistent lifecycles (per tool call vs. per conversation vs. global),
+divergent behavior across transports, and application state coupled to the
+connection, which breaks under load balancing and rolling upgrades.
+
+After SEP-2575 (stateless by default) and SEP-2322 (multi round-trip requests)
+removed the transport-level need for sessions, the core maintainers
+[voted][vote-2536] to remove sessions from the protocol entirely (see the
+[decision doc][decision-doc]). The recommended replacement is the **explicit
+state handle** pattern: servers return an opaque ID from a tool call
+(`create_basket() -> basket_id`) and clients pass it back as a tool argument in
+later calls.
+
+Since then, the community has pushed back with scenarios where an identifier
+that spans a sequence of requests still seems necessary (see [PR
+#2822][pr-2822]). This document frames what we would need to decide to
+reintroduce such a concept, likely first as an experimental extension.
+
+[vote-2536]:
+    https://github.com/modelcontextprotocol/modelcontextprotocol/discussions/2536
+[decision-doc]: ./sessions-vs-sessionless-decision.md
+[pr-2822]:
+    https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2822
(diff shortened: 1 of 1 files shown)


docs/sessions-reintroduction-problem-statement.md
```diff
@@ -0,0 +1,361 @@
...
+    https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2822#issuecomment-4625554258
+[hellboy-routing]:
+    https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2822#issuecomment-4605717954
+
+### 2. Observability and audit correlation
+
+**TL;DR:** There is no standard field to correlate the requests of one logical
+unit of work in logs and audit trails; every deployment invents its own.
+
+_Mentioned by @javapro108 ([comment][javapro108-observability])._
```
evalstate, line 147: I would like this too

docs/sessions-reintroduction-problem-statement.md
```diff
@@ -0,0 +1,361 @@
...
+needs bespoke configuration; off-the-shelf host applications won't set or
+forward nonstandard headers; and it only works when one party owns both ends of
+the wire, which MCP customers generally don't.
+
+##### Option B: An initialization tool that mints a state handle
+
+The server exposes a tool the client must call first (`init_conversation() ->
+handle`); the handle is passed on every subsequent call and, via
+[SEP-2243][sep-2243]'s `x-mcp-header` annotation, mirrored into an
+`Mcp-Param-{name}` header a gateway can route on without parsing bodies:
```
LucaButBoring, line 117: FWIW this means gateways need to reshape the tools behind them to add a `handle` field that the upstream server doesn't declare (the upstream server isn't aware of the gateway). That's likely to conflict with servers that actually already use a handle for their own state independently of the gateway itself. In the general case, this seems difficult to get right.

docs/sessions-reintroduction-problem-statement.md
```diff
@@ -0,0 +1,361 @@
...
+##### Option A: Non-MCP fix — existing telemetry standards
+
+The observability stack already has most of the pieces. The OpenTelemetry
+GenAI semantic conventions reserve the exact attribute this category wants:
+[`gen_ai.conversation.id`][otel-genai-attrs], "the unique identifier for a
+conversation (session, thread)." And MCP has a documented carrier for OTel
+context: [SEP-414][sep-414] reserves `traceparent`, `tracestate`, and `baggage`
+in `_meta`, on every transport including stdio. Assembled, a host could carry a
+conversation ID in W3C Baggage via `_meta`, and servers could record it as
+`gen_ai.conversation.id`:
```
LucaButBoring, line 178: Feels like this runs into all the host divergence questions of sessions (e.g. what is a conversation ID in Anthropic's remote MCP API?)

docs/sessions-reintroduction-problem-statement.md
```diff
@@ -0,0 +1,361 @@
...
+```
+initialize               [X-Conv-Id: conv-7]  -> cold start: provision microVM A
+tools/call: query(...)   [X-Conv-Id: conv-7]  -> routed to warm microVM A
+tools/call: submit(...)  [X-Conv-Id: conv-7]  -> routed to warm microVM A
+```
+
+The limits: nothing standardizes the header name, so every host/gateway pair
+needs bespoke configuration; off-the-shelf host applications won't set or
+forward nonstandard headers; and it only works when one party owns both ends of
+the wire, which MCP customers generally don't.
```
LucaButBoring, line 110: This is itself a nonstarter - any sort of association ID can't be a bespoke negotiation, or it only works if you explicitly build a host application to support a specific gateway.

ryan-s-roberts:
There is possibly an additional category here for identifier misattribution and loss. The existing categories address what the identifier is for, but don't address what happens if it's wrong. A relayed identifier that is absent, stale, or belonging to a peer conversation produces no error and no server-side signal, and (handle, auth_context) cannot detect it when both conversations belong to the same principal.

Does the WG consider this a use case, or an implementation concern?

PR #50, opened 2026-07-23
docs: add upcoming transport WG tracks roadmap

kurtisvg:
Strawman proposal for roadmap for the next ~6 months.

### DIFF
--- a/docs/upcoming-tracks-roadmap.md
+++ b/docs/upcoming-tracks-roadmap.md
@@ -1,0 +1,342 @@
+# Transports WG: Upcoming Tracks Roadmap
+
+This document is a high-level outline of the tracks the Transports Working Group
+expects to take up next. Each track gets a short brief: the goal (what problem
+it solves), the core idea of the solution, key requirements, and the open
+questions we still need to answer. As tracks mature, each brief should graduate
+into a full problem statement or proposal under `proposals/`.
+
+Tracks are listed in rough priority order.
+
+Related work already in flight:
+
+* [Multi Round-Trip Requests:
+  Brief](../proposals/multi-round-trip-requests-track-brief.md)
+* [Sessions Reintroduction Problem
+  Statement](sessions-reintroduction-problem-statement.md)
+* Caching & Optimization Track (TTLs and ETags for resources and list responses)
+  — several tracks below interact with it and are cross-referenced.
+
+---
+
+## 1. Tool Versioning
+
+**Goal:** Eliminate skew between the tool definition a client has cached and the
+tool the server actually executes. As caching (TTLs/ETags) makes long-lived
+cached `tools/list` results normal, clients need a way to know *which version*
+of a tool they are calling, and servers need a way to detect calls made against
+stale definitions.
+
+**Core Idea:** Carry the ETag mechanism from the Caching & Optimization track
+down to the individual tool definition. Each tool in `tools/list` gets an opaque
+version identifier; clients echo it on `tools/call`; the server can then detect
+a mismatch and respond with a standard "stale tool definition" error that tells
+the client to refresh, rather than silently executing with different semantics
+than the model saw.
+
+**Key Requirements:**
+
+* A standard per-tool version field in tool definitions (opaque, like an ETag —
+  not required to be ordered or semantic).
+* An echo mechanism on `tools/call` (likely `_meta`) carrying the version the
+  client believes it is calling.
+* Defined mismatch semantics: a standard error code/shape that signals "re-fetch
+  the tool list and retry", distinguishable from tool execution errors.
+* Coherent layering with list-level caching: list ETag/TTL governs when to
+  re-fetch; per-tool versions govern call-time validity.
+
+**Open Questions:**
+
+* Opaque hash vs. semantic version? An opaque hash is simplest and matches ETag
+  semantics; semver would let servers express compatible vs. breaking changes
+  but invites misuse.
+* Is echoing the version on `tools/call` required or optional? Optional
+  preserves backwards compatibility but weakens the guarantee.
+* Mismatch behavior: hard failure, or server MAY accept if the change is
+  compatible? Who decides compatibility?
+* What exactly is hashed — the whole definition (description edits invalidate)
+  or just the machine-relevant parts (schema, annotations)? Note interaction
+  with i18n: localized descriptions must not change the version, or every
+  language gets a different one.
+* Does the same mechanism generalize to prompts and resources?
+
+---
+
+## 2. Pluggable Transports
+
+**Goal:** Allow MCP to run over transports beyond stdio and Streamable HTTP
+(e.g., gRPC, WebSockets, in-process/embedded) without forking SDKs or
+re-specifying protocol behavior per transport. Organizations with existing RPC
+infrastructure should be able to carry MCP over it while application code stays
+unchanged.
+
+**Core Idea:** Separate protocol semantics from transport bindings. Define an
+abstract transport contract — the guarantees any conforming transport must
+provide (message framing, ordering, request/response correlation,
+bidirectional/server-initiated messages, a metadata channel) — and have SDKs
+expose that contract as a stable interface. Custom transports plug in beneath
+the SDK; the spec describes required semantics rather than a single wire format.
+
+**Key Requirements:**
+
+* A formal transport contract: what a transport MUST provide (delivery,
+  ordering, duplex messaging, metadata/headers) and what is optional.
+* Stable, consistent SDK transport interfaces across languages, so a transport
+  implementation written against the contract works with unmodified
+  server/client application code.
+* A way for the protocol layer to detect optional transport features (e.g.,
+  server push, resumability, metadata channel) and degrade gracefully.
+* Guidance for mapping the MCP lifecycle (initialization, shutdown,
+  reconnection) onto a new transport.
+* Ideally, a conformance checklist or test suite for third-party transports.
+
+**Open Questions:**
+
+* Where is the line between transport-defined and protocol-defined behavior?
+  Sessions, resumability, and auth currently lean on HTTP-specific mechanisms.
+* How do we handle transports without a natural metadata/header channel —
+  require `_meta` mirroring everywhere, or make headers optional?
+* Do we need a naming/registry scheme for transports (URI schemes, capability
+  identifiers) so clients and servers can agree on one?
+* Relationship to the HTTP-over-stdio track: if HTTP semantics become the
+  canonical binding, does "pluggable transport" reduce to "anything that can
+  carry HTTP", or do we keep a lower-level message contract?
+* How do we recover the previously deferred pluggable transports PR and
+  reconcile it with the unified proposal (see 2026-05-27 meeting notes)?
+
+---
+
+## 3. HTTP over STDIO
+
+**Goal:** Reduce the behavioral divergence between the stdio and Streamable HTTP
+transports. Today stdio has a pile of bespoke behavior (framing, lifecycle, no
+headers), which means every HTTP-native feature — headers, status codes,
+caching, compression, auth, language negotiation — needs a second,
+stdio-specific design or simply doesn't work locally.
+
+**Core Idea:** Instead of maintaining two transport designs, run HTTP itself
+over the stdin/stdout pipe: the subprocess speaks the same Streamable HTTP
+binding it would speak over a socket, just with the byte stream attached to
+pipes. stdio becomes "a connection like any other", and the spec converges on a
+single transport binding with one behavior model.
+
+**Key Requirements:**
+
+* A framing choice (HTTP/1.1 vs. HTTP/2 over the pipe) that supports concurrent
+  in-flight requests and server-initiated messages — plain HTTP/1.1
+  request/response over a single pipe pair is not enough.
+* Full parity for metadata: headers must work identically to network HTTP so
+  features like i18n (`Accept-Language`) and caching (`ETag`) apply unchanged.
+* A detection/negotiation story: clients must be able to tell whether a spawned
+  server speaks legacy newline-delimited JSON-RPC or HTTP-over-stdio, with a
+  long deprecation path for the former.
+* Implementation burden must stay small — local stdio servers are often tiny
+  scripts; requiring a full HTTP/2 stack may be prohibitive outside major SDK
+  languages.
+* Compatibility with the ongoing stdio process-lifecycle work (crash-only
+  restarts, stderr conventions).
+
+**Open Questions:**
+
+* HTTP/1.1 with a bidirectional convention vs. HTTP/2 multiplexing? HTTP/2 gives
+  real streams but a much heavier dependency.
+* How do server-initiated requests map onto HTTP's client-initiated model — a
+  reverse HTTP channel over the same pipes, or SSE streams as in Streamable
+  HTTP?
+* Does this actually reduce complexity, or relocate it (an HTTP parser in every
+  hello-world stdio server)? What does the minimal conforming server look like?
+* Is this a new named transport, a stdio v2 revision, or an optional binding
+  negotiated at startup?
+* Does auth over local pipes inherit HTTP auth semantics, or is it explicitly
+  out of scope as it is for stdio today?
+
+---
+
+## 4. Streaming
+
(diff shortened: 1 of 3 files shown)


docs/upcoming-tracks-roadmap.md
```diff
@@ -0,0 +1,342 @@
...
+  semantics; semver would let servers express compatible vs. breaking changes
+  but invites misuse.
+* Is echoing the version on `tools/call` required or optional? Optional
+  preserves backwards compatibility but weakens the guarantee.
+* Mismatch behavior: hard failure, or server MAY accept if the change is
+  compatible? Who decides compatibility?
+* What exactly is hashed — the whole definition (description edits invalidate)
+  or just the machine-relevant parts (schema, annotations)? Note interaction
+  with i18n: localized descriptions must not change the version, or every
+  language gets a different one.
```
SamMorrowDrums: Shaun and I discussed that this is a feature, were changing the language and not relisting tools would cause the mismatch that would then tell the client it needs to.

docs/upcoming-tracks-roadmap.md
```diff
@@ -0,0 +1,342 @@
...
+
+## 1. Tool Versioning
+
+**Goal:** Eliminate skew between the tool definition a client has cached and the
+tool the server actually executes. As caching (TTLs/ETags) makes long-lived
+cached `tools/list` results normal, clients need a way to know *which version*
+of a tool they are calling, and servers need a way to detect calls made against
+stale definitions.
+
+**Core Idea:** Carry the ETag mechanism from the Caching & Optimization track
```
SamMorrowDrums: ETAGs are also about the response content traditionally - which works for the list responses but not for the tool calls, which need to be a custom approach.

docs/upcoming-tracks-roadmap.md
```diff
@@ -0,0 +1,342 @@
...
+
+**Open Questions:**
+
+* Where is the line between transport-defined and protocol-defined behavior?
+  Sessions, resumability, and auth currently lean on HTTP-specific mechanisms.
+* How do we handle transports without a natural metadata/header channel —
+  require `_meta` mirroring everywhere, or make headers optional?
+* Do we need a naming/registry scheme for transports (URI schemes, capability
+  identifiers) so clients and servers can agree on one?
+* Relationship to the HTTP-over-stdio track: if HTTP semantics become the
```
SamMorrowDrums: At very least I think there's a case to be made for envelope that carries the stuff in the http spec, that doesn't have to involve http.

SamMorrowDrums: Special case of http header auth handling is the only scary part here - stuff that has built in handling.

docs/upcoming-tracks-roadmap.md
```diff
@@ -0,0 +1,342 @@
...
+  real streams but a much heavier dependency.
+* How do server-initiated requests map onto HTTP's client-initiated model — a
+  reverse HTTP channel over the same pipes, or SSE streams as in Streamable
+  HTTP?
+* Does this actually reduce complexity, or relocate it (an HTTP parser in every
+  hello-world stdio server)? What does the minimal conforming server look like?
+* Is this a new named transport, a stdio v2 revision, or an optional binding
+  negotiated at startup?
+* Does auth over local pipes inherit HTTP auth semantics, or is it explicitly
+  out of scope as it is for stdio today?
```
SamMorrowDrums: The auth spec still has challenges here - protected resource metadata, and identity of the server conveyed to auth server doesn't entirely make sense.

docs/upcoming-tracks-roadmap.md
```diff
@@ -0,0 +1,342 @@
...
+
+**Open Questions:**
+
+* Is streaming a new mechanism or an extension of tasks? The WG flagged that
+  auto-streaming may conflict with existing task cancellation cases; simplifying
+  tasks may be in scope.
+* Partial results on error: is a stream that fails midway a failed call, a
+  partial success, or client's choice?
+* What guidance do clients get on spooling vs. forwarding chunks to the model as
+  they arrive?
```
SamMorrowDrums:
```suggestion
* What guidance do clients get on spooling vs. forwarding chunks to the model as
  they arrive?
* what guidance do clients get on redirecting to RAG pipelines/files vs straight to model context
```

docs/upcoming-tracks-roadmap.md
```diff
@@ -0,0 +1,342 @@
...
+with thousands of rows, generated images, long documents — as a stream of chunks
+instead of a single terminal response. Today the entire result must be buffered
+on both sides and the client sees nothing until the call completes.
+
+**Core Idea:** Extend tool results (and potentially other responses) with
+partial/chunked delivery over the existing streaming machinery: a call yields an
+ordered sequence of result chunks followed by a terminal message with final
+status. Chunking is content-block oriented so both structured data and binary
+content work, and the semantics are transport-independent (SSE carries it
+naturally; stdio needs the same model — see the HTTP-over-stdio track).
```
SamMorrowDrums: Maybe it's a an open question, but shouldn't this just be an improvement of Resources, and tools could return a streaming resource link, but that Resources should possibly still be the mechanism too.

I'm saying this because I think this kind of replaces Resources conceptually in some ways.

docs/upcoming-tracks-roadmap.md
```diff
@@ -0,0 +1,342 @@
...
+
+**Key Requirements:**
+
+* Chunk framing with ordering guarantees and a distinguishable terminal message
+  carrying final status.
+* Cancellation mid-stream, in both directions, with defined cleanup semantics.
+* A clear boundary with tasks: when should a long-running call be a task vs. a
+  stream, and how does streaming a *task's* output work? (Raised in the
+  2026-06-24 meeting; cancellation interaction is the sharp edge.)
+* Resumability: what happens when a stream is interrupted — redelivery from an
```
SamMorrowDrums: This could easily end up adding another dependency on http if we're not careful, and not using http features for this risks making it more expensive than it needs to be, so definitely need to consider this.

docs/upcoming-tracks-roadmap.md
```diff
@@ -0,0 +1,342 @@
...
+* How does "scope" language square with a stateless protocol? The
+  refresh-obligation framing may replace scope language entirely.
+* Interaction with the caching track: capability-dependent lists make the
+  capability set a cache key — how do list ETags/TTLs account for it?
+* Can a client under-declare and upgrade later (progressive disclosure), and
+  what does that do to cached lists?
+
+---
+
+## 7. Sessions
```
SamMorrowDrums: I think we should sit on this - it's good to let the handling using the new spec and possible extensions people propose etc. to let it bake before hand.

docs/upcoming-tracks-roadmap.md
```diff
@@ -0,0 +1,342 @@
...
+  refresh cached lists (reframing suggested in review: this is a client-refresh
+  responsibility, not a server-persistence assumption, keeping the protocol
+  stateless).
+* Privacy by design: clients must not be forced to disclose their full
+  capability set to arbitrary servers — capability disclosure is a
+  fingerprinting vector (raised in PR review, with the HTTP user-agent history
+  as the cautionary tale).
+* No requirement for servers to remember per-client capabilities across
+  requests.
+
```
SamMorrowDrums: SDKs may not know if a user is using them except during runtime, so they could be wrong even if provided. Not saying that's a blocker, just a a challenge in making it a contract and not just advice.

docs/upcoming-tracks-roadmap.md
SamMorrowDrums: Also distributed tracing might be a separate extension proposal, that could cut across more protocols than MCP.

docs/upcoming-tracks-roadmap.md
```diff
@@ -0,0 +1,342 @@
...
+  mandatory?
+* Interaction with the caching track: language becomes a cache key (`Vary:
+  Accept-Language`); list ETags and tool versions must not vary by language, or
+  must account for it explicitly.
+* Translating model-facing text changes agent behavior — do we need guidance on
+  evaluating tool-selection quality across languages?
+
+---
+
+## 6. Capabilities for Tools
```
gjz22: @kurtisvg this is an alternative wording for this:

6. Capabilities for Tools                                                                                                                              
                                                            
  Goal: Resolve the cross-request coordination gap from SEP-2575: without                                                                                
  additional mechanism, neither the server nor the client knows at tools/list
  time whether a tool will actually work given the client's capabilities, leading                                                                        
  to raw capability errors reaching end users and wasted tokens as the model plans                                                                       
  toward tools it cannot run.                                                                                                                            
  (Tracked in https://github.com/modelcontextprotocol/transports-wg/pull/31.)                                                                            
                                                            
  Core Idea: Servers annotate each tool with hard capability requirements via                                                                            
  io.modelcontextprotocol/requiredCapabilities in the tool's _meta (same
  shape as ClientCapabilities, covering core and extension capabilities).                                                                                
  Clients match their own capabilities against annotations and decide how to                                                                             
  respond -- filter, disable, or allow-and-error. The server returns the same
  annotated tool list to every client. Tools that can operate with or without a                                                                          
  capability ("hybrid mode") need no annotation.                                                                                                         
                                                                                                                                                         
  Server validation at call time is the enforcement backstop: tools/call returns                                                                         
  MissingRequiredClientCapabilityError (-32021) if a required capability is
  absent.                                                                                                                                                
   
  Key Requirements:                                                                                                                                      
                                                            
  - Client-side filtering is advisory, call-time server validation is mandatory.
  - Tool lists are identical for all clients and cacheable without the capability                                                                        
  set as a cache key, simplifying proxy servers and aligning with SEP-2549 (TTL
  for List Results).                                                                                                                                     
  - No cross-request capability declarations and no server-side storage of
  per-client state.                                                                                                                                      
                                                            
  Open Questions:


  - Client-side vs. server-side filtering -- the SEP adopts client-side with a                                                                           
  comparison matrix. Server-side (via related-request capabilities) enables
  dynamic substitution and complex adaptation but makes tool lists                                                                                       
  capability-dependent, complicating caching and removing the client's ability
  to present unavailable tools.                                                                                                                          
  - Substitution: hybrid mode covers most cases, but distinct fallback tools
  cannot be statically linked today. A whenCapabilitiesUnavailable annotation                                                                            
  was excluded for complexity -- revisit if practice shows the need?
  - Can recursive sub-field matching (e.g., requiring sampling.tools                                                                                     
  specifically) be reliably implemented across all clients, including for
  experimental and future extension capabilities?                                                                                                        
  - Interaction with the caching track: the current proposal keeps tool lists
  capability-independent and trivially cacheable, but if the approach shifts to                                                                          
  server-side filtering, the capability set becomes a cache key -- how do list
  ETags/TTLs account for it? Relatedly, can a client under-declare capabilities                                                                          
  and upgrade later (progressive disclosure), and what does that do to cached                                                                            
  tool lists?

docs/upcoming-tracks-roadmap.md
```diff
@@ -0,0 +1,342 @@
...
+**Goal:** Eliminate skew between the tool definition a client has cached and the
+tool the server actually executes. As caching (TTLs/ETags) makes long-lived
+cached `tools/list` results normal, clients need a way to know *which version*
+of a tool they are calling, and servers need a way to detect calls made against
+stale definitions.
+
+**Core Idea:** Carry the ETag mechanism from the Caching & Optimization track
+down to the individual tool definition. Each tool in `tools/list` gets an opaque
+version identifier; clients echo it on `tools/call`; the server can then detect
+a mismatch and respond with a standard "stale tool definition" error that tells
```
gjz22: There's also the use case where there is a migration period from older versions to newer versions that the client can handle. For example, maybe the client has hooks on a stale definition that modify the parameter inputs given to the tool in some way. The next definition changes those parameter names and now the hooks are invalid and the client breaks.

Therefore the server MAY respond with an error but it doesn't have to. The client can then use the ETag mismatch in metadata in the returned tool call to kick off a process (automated or partially manual) to ensure they can migrate to the latest version without issue.

Also, tools/list should be able to respond based on an ETag as well, where older tool definitions are supported. This maybe gets into what the mechanism should be, for example, should a whole server have an ETag that is consistent across tools or each tool has an ETag? There are tradeoffs to either which maybe are better addressed in the SEP

docs/upcoming-tracks-roadmap.md
```diff
@@ -0,0 +1,413 @@
...
+  mandatory?
+* Interaction with the caching track: language becomes a cache key (`Vary:
+  Accept-Language`); list ETags and tool versions must not vary by language, or
+  must account for it explicitly.
+* Translating model-facing text changes agent behavior — do we need guidance on
+  evaluating tool-selection quality across languages?
+
+---
+
+## 7. Capabilities for Tools
```
gjz22: Suggested change for this section:

6. Capabilities for Tools                                                                                                                              
                                                                                                                                                       
  Goal: Resolve the cross-request coordination gap from SEP-2575: without                                                                                
  additional mechanism, neither the server nor the client knows at tools/list
  time whether a tool will actually work given the client's capabilities, leading                                                                        
  to raw capability errors reaching end users and wasted tokens as the model plans
  toward tools it cannot run.                                                                                                                            
  (Tracked in https://github.com/modelcontextprotocol/transports-wg/pull/31.)
                                                                                                                                                         
  Core Idea: Servers annotate each tool with hard capability requirements via
  io.modelcontextprotocol/requiredCapabilities in the tool's _meta (same                                                                                 
  shape as ClientCapabilities, covering core and extension capabilities).                                                                                
  Clients match their own capabilities against annotations and decide how to
  respond -- filter, disable, or allow-and-error. The server returns the same                                                                            
  annotated tool list to every client. Tools that can operate with or without a
  capability ("hybrid mode") need no annotation.                                                                                                         
   
  Server validation at call time is the enforcement backstop: tools/call returns                                                                         
  MissingRequiredClientCapabilityError (-32021) if a required capability is
  absent.                                                                                                                                                
   
  Key Requirements:                                                                                                                                      
                                                            
  - Tool requirements are machine-readable in tools/list; client-side filtering
  is advisory, call-time server validation is mandatory.
  - Tool lists are identical for all clients and cacheable without the capability
  set as a cache key, simplifying proxy servers and aligning with SEP-2549 (TTL
  for List Results).                                                                                                                                     
  - No cross-request capability declarations and no server-side storage of
  per-client state.                                                                                                                                      
                                                            
  Open Questions:

  - Client-side vs. server-side filtering -- the SEP adopts client-side with a                                                                           
  comparison matrix. Server-side (via related-request capabilities) enables
  dynamic substitution and complex adaptation but makes tool lists                                                                                       
  capability-dependent, complicating caching and removing the client's ability
  to present unavailable tools.
  - Substitution: hybrid mode covers most cases, but distinct fallback tools                                                                             
  cannot be statically linked today. A whenCapabilitiesUnavailable annotation
  was excluded for complexity -- revisit if practice shows the need?                                                                                     
  - Can recursive sub-field matching (e.g., requiring sampling.tools
  specifically) be reliably implemented across all clients, including for                                                                                
  experimental and future extension capabilities?
  - Interaction with caching: the current proposal keeps tool lists                                                                                      
  capability-independent and trivially cacheable, but if the approach shifts to
  server-side filtering, the capability set becomes a cache key -- how do list
  ETags/TTLs account for it? Relatedly, can a client under-declare capabilities                                                                          
  and upgrade later (progressive disclosure), and what does that do to cached
  tool lists?

docs/upcoming-tracks-roadmap.md
kurtisvg: Rewording to say "implementing an etags-like" mechanism

docs/upcoming-tracks-roadmap.md
kurtisvg: I think "tool result audience" might be an orthogonal problem that's being solved as part of the core primitives wg

docs/upcoming-tracks-roadmap.md
kurtisvg: My 2c is that resources and tool results seem like two separate concepts. There is some work going on in file transport wg to do file and resource streaming. We should be in the loop but not necessarily the same effort.

SamMorrowDrums (approved):

evalstate (approved):

PR #53, opened 2026-08-20
Proposal: Pluggable Transport Interface for MCP SDKs

kziemski:
## Summary

Standardizes a public, replaceable **transport extension point** across official MCP SDKs: the SDK boundary between MCP protocol processing and the mechanism that carries MCP JSON-RPC messages. A conforming SDK exposes a public transport abstraction through which a third-party transport can send and receive complete MCP JSON-RPC messages without forking or modifying protocol dispatch.

## Scope

- **In**: JSON-RPC-preserving custom transports (WebSocket, `postMessage`/`MessagePort`, Unix sockets, named pipes, gRPC/Cap'n Web tunnels carrying opaque MCP JSON-RPC, in-process channels).
- **Out (deliberately)**: native RPC bindings that replace the JSON-RPC method layer (e.g. typed protobuf gRPC services, native Cap'n Web object-capability mappings) — deferred to a separate proposal.
- **Out**: adding any new standard MCP transport binding.

## Why this framing

Narrows the deferred [SEP-2598](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2598): keeps MCP JSON-RPC as the semantic boundary (which the 2026-07-28 spec normatively requires for custom transports), standardizes the SDK extension seam behaviorally rather than by identical API, and centralizes conformance vectors instead of requiring every Tier-1 SDK to maintain its own harness.

## What's here

- `proposals/XXXX-pluggable-transports.md` — the proposal (draft).
- `docs/pluggable/` — supporting research: cross-SDK transport survey and RPC-transport analysis.

## Open for discussion

Draft for WG comment. Highlights for review:

- Message-oriented boundary (one complete JSON-RPC message) vs exchange-oriented API — §9.
- Opaque Transport Context as the only carrier-specific escape hatch — §5.
- Behavioral conformance; concurrency/correlation as the one MUST — §Conformance.
- SDK alignment table and per-SDK guidance — §SDK Alignment.

Ready for feedback.

### DIFF
--- a/docs/pluggable/pluggable-transports-interface-research.md
+++ b/docs/pluggable/pluggable-transports-interface-research.md
@@ -1,0 +1,1023 @@
+# Pluggable MCP Transports Across the Official SDKs
+
+## Executive assessment
+
+**Yes. A narrower “pluggable transport” step—keeping MCP messages as JSON-RPC while making the transport/binding replaceable—is both feasible and, after the July 28, 2026 MCP specification, much better aligned with the protocol than SEP-2598 was when it was proposed.** The current specification now explicitly defines a transport as a **binding** that determines framing, delivery, metadata carriage, cancellation, and termination while leaving MCP message semantics unchanged. It explicitly permits custom transports, but requires them to preserve the JSON-RPC message format, MCP message patterns, and the per-request metadata model.
+
+That distinction is the key architectural answer to this research:
+
+> **MCP core semantics → JSON-RPC message model → pluggable binding/transport → actual communication substrate**
+
+`stdio` and Streamable HTTP are simply the two standardized bindings today. A WebSocket, `MessagePort`/`postMessage`, TCP socket, Unix socket, RPC framework, or another communication substrate can occupy the layer beneath the JSON-RPC message model without redefining MCP itself. The specification now says almost exactly this: custom transports may use any channel supporting bidirectional message exchange, while reliable bidirectional byte streams should normally reuse stdio's newline-delimited JSON-RPC framing.
+
+The SDK review produces an unusually strong result: **most of the ten official SDKs already have either an explicit transport interface/protocol or a public transport-injection seam.** Go, Java, C#, PHP, Rust, and Swift expose particularly clear abstractions; Kotlin already ships a WebSocket transport and an in-memory `ChannelTransport`; Python v2 accepts a custom `Transport` directly; TypeScript has public transport injection and a sophisticated internal transport split; Ruby has a client-side custom-transport contract but is less cleanly abstracted on the server side.
+
+My resulting feasibility assessment is:
+
+| SDK | Existing pluggability | Work to establish a stable transport SPI | Assessment |
+|---|---|---:|---|
+| TypeScript | Public `connect(transport)`; stdio/HTTP/in-memory; modern HTTP internally uses a per-request transport | Clarify/stabilize the v2 SPI and reconcile connection-oriented vs per-exchange APIs | **Low–Medium** |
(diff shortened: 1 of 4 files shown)


docs/pluggable/pluggable-transports-interface-research.md
```diff
@@ -0,0 +1,1023 @@
...
+
+Those are separable ideas, and the second one generated much of the design resistance.
+
+Review discussion raised questions such as whether a native gRPC implementation should need a JSON-RPC request ID at all, whether all transports should support all MCP notifications, whether `initialize` should exist in a stateless transport, and whether response correlation should be exposed through request IDs in a general transport API.
+
+Another objection concerned process rather than architecture: requiring every Tier-1 SDK to expose an interface **and** maintain its own reusable conformance harness would impose substantial cross-SDK maintenance and tier-review burden. Reviewers suggested treating transport support more like an optional extension and putting transport-specific tests in the common conformance ecosystem instead. The SEP also still lacked its required prototype implementation.
+
+Those objections point toward a much better first SEP:
+
+> **Do not change MCP's JSON-RPC representation. Standardize the SDK extension seam for custom bindings that carry existing MCP JSON-RPC.**
```
kurtisvg, line 97: Part of the problem not entirely captured here is that the JSON-RPC representation is part of the problem. e.g. the duplication of info between headers and _meta is because the current data model is overly focused on the transport. 

In addition, there are some aspects that are not represented (e.g. auth)

docs/pluggable/pluggable-transports-interface-research.md
```diff
@@ -0,0 +1,1023 @@
...
+
+That eliminates most of the semantic translation problem.
+
+There is an additional reason not to revive the transcoding part of SEP-2598 unchanged: **the final July 28, 2026 specification now normatively says custom transports must preserve the JSON-RPC message format.**
```
kurtisvg, line 101: This implication is specifically message format right? So we are saying that custom transports MUST preserve the JSON-RPC message format on the wire?

kurtisvg (commented):
One thing I think this is lacking is strong motivation on why pluggable transports are desired. This talks a lot about the current work done in existing SDKs, but doesn't really describe the motivation for such a thing.

docs/pluggable/pluggable-transports-interface-research.md
kziemski: Its whats in the spec and we gain alot of benefit in moving pluggable transports forward without simultaneously forcing a SEP touching that livewire. Spec's says pluggable transports we need an alignment on a plug. once we have that we can continue bring the sdk's to conformance and best practices.

docs/pluggable/pluggable-transports-interface-research.md
kziemski: This is a proposal to bring forward the work on conformance to the spec that pluggable transports. Auth would be a critical component and any related PR work would require not breaking conformance for the existing sdks. I would progress by building PR's that provide the abstraction while not breaking conformance for essential capabilites like auth.

docs/pluggable/pluggable-transports-interface-research.md
kziemski: With that aligned abstraction it would be easier to put forward the capability to weave in alternatives while not altering message format for core transports

kziemski:
> One thing I think this is lacking is strong motivation on why pluggable transports are desired. This talks a lot about the current work done in existing SDKs, but doesn't really describe the motivation for such a thing.

I can only offer analysis of why i think each brought each of those projects into existence.  
I didn't create them so their intention was their own.  
For the native/in-process rpc transports that were created here it was crossing a boundary between client and server where stdio and http would not be possible without keeping it completely out of the environment. capnweb, rpc, cross frame (postmessage). 

There are also regulated industries that have strict boundaries that make streamablehttp/stdio not practical.  

Pursuing this avenue first will enable their use cases, address the issues and keep them in the ecosystem but outside the spec.

